PR #157, opened 2016-02-24
Update README.md

nuclearace:
Kitura net version changed


dfirsht:
In the master branch, we still want to use 0.2 as 0.3 is the development version. If you want to open this PR against the development branch, we'll accept it there.

Issue #19, opened 2016-02-08
Move from the forked version of libdispatch (maintained by Chris) and use the official one from apple


dfirsht:
<a href="https://github.ibm.com/rfdickerson"><img src="https://avatars.github.ibm.com/u/1506?" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [rfdickerson](https://github.ibm.com/rfdickerson)**
_Thursday Jan 07, 2016 at 21:48 GMT_
_Originally opened as https://github.ibm.com/ibmswift/Phoenix/issues/19_

----

Currently, the swiftc cannot interpret the proprocessing macros inside of header files such as #define's. For example, I can build my program and builds and can execute, 

```swift
import CDispatch

let q = dispatch_get_main_queue()
let t = dispatch_time(DISPATCH_TIME_NOW, Int64(5*NSEC_PER_SEC))

dispatch_after(t, q, {
	print("hello")
})

dispatch_main()

```

The Makefile is:

```Makefile
DISPATCH_LIBRARY=-Xcc -fblocks -ldispatch -lpthread -Xlinker -lBlocksRuntime -I Packages/CDispatch-1.0.2

all: main.o

main.o: main.swift
	swiftc main.swift ${DISPATCH_LIBRARY}


clean:
	rm *.o
```

The executable runs, waits 5 seconds, and prints hello.

However, if I try to use 

DISPATCH_QUEUE_CONCURRENT

The Swift compiler cannot resolve that symbol. This symbol is really a #define, and I suspect the Swift frontend compiler cannot resolve the identifier in any typesafe way since it's just a simple code-replacement. 

```C
#define DISPATCH_QUEUE_CONCURRENT \
		DISPATCH_GLOBAL_OBJECT(dispatch_queue_attr_t, \
		_dispatch_queue_attr_concurrent)
```

A proposed solution to this problem would be to add to a custom header file a new definition of DISPATCH_QUEUE_CONCURRENT to instead be a constant global variable instead of a C preprocessor directive.

dfirsht:
Now that the PR is accepted, we should confirm that the official libdispatch library works with Kitura

dfirsht:
I am moving this to blocked again as I am unable to figure out how to get libdispatch properly installed and configured. I configured libdispatch with `./configure --with-swift-toolchain=<path>` and then made it but am not sure what to do next. I am waiting on a response from @dgrove-oss or anyone else on the runtime team on how to do this.

[...]

PR #173, opened 2016-02-26
Fix vagrant provisioning by running as unprivileged user

cjwirth:
After the https://github.com/IBM-Swift/Kitura/commit/c30c088d5980d7932fb97c1c57f03506ab926486 commit, which solves [#19 Move from the forked version of libdispatch (maintained by Chris) and use the official one from apple](https://github.com/IBM-Swift/Kitura/issues/19), the vagrant provisioning script stopped working. 

The main reason being that it was still pointing towards the fork of libdispatch, and the modulemap file it was pointing at no longer existing.

I went in and updated the script to work with the new version of it, and I have been able to run the sample app ("You're running Kitura") from just

~~~
> vagrant up
> vagrant ssh
> Kitura/.build/debug/KituraSample
~~~

### DIFF
--- a/vagrantfile
+++ b/vagrantfile
@@ -27,7 +27,7 @@ Vagrant.configure(2) do |config|
     s.inline = "sudo sed -i '/tty/!s/mesg n/tty -s \\&\\& mesg n/' /root/.profile"
   end
 
-  config.vm.provision "shell", inline: <<-SHELL
+  config.vm.provision "shell", privileged: false, inline: <<-SHELL
 
 ### Install packages
 # 1. Install compiler, autotools


cjwirth:
As a follow up question, everything that the provisioning installs (the swift development snapshot, pcre, as well as the Kitura repository itself) are all owned by `root`, rather than `vagrant`. 

Is this intentional? If so, what is the reasoning behind it?

kweinmeister:
Thanks for your PR! And glad you were able to get the build running successfully with Vagrant.

A couple minor things I'd recommend changing in the PR: removing the constant for the module map at the top of the file, and removing the word "patched" from the comment about downloading the patched libdispatched. Also, it wasn't intentional to install as root; that's just the default. We could do something like the following?
   config.vm.provision "shell", privileged: false, inline: <<-SHELL

cjwirth:
Ah, nice catches. I've went in and fixed them, so take a look.

I think it's better to not install as root, so I've changed that as well. Yep, running the script as unprivileged worked :smile:

kweinmeister:
Awesome! Can you follow the process in [CONTRIBUTING.md](/IBM-Swift/Kitura/blob/master/CONTRIBUTING.md), and then we can move forward?

cjwirth:
Hey, sorry for the delay. Was busy putting on a conference.

It looks like most of the changes were already put into the develop branch (thanks!). 
I rebased, and the only change that remained was running as an unprivileged user :sweat_smile:

cjwirth:
I've also made an issue https://github.com/IBM-Swift/Kitura/issues/232 in order to conform with [CONTRIBUTING.md](https://github.com/IBM-Swift/Kitura/blob/master/CONTRIBUTING.md) even though this has become a one line change :sweat_smile:

dfirsht:
@cjwirth Hey, sorry for the delay on getting this accepted. Have you sent in your signed CLA yet? Unfortunately we cannot accept any contributions from you until you do so.

cjwirth:
Yeah I sent it to the email that was listed inside the pdf. Let me know if something is off? 
image

dfirsht:
@cjwirth No problem, it just wasn't processed yet. We can merge it now.

Issue #281, opened 2016-03-15
swift build --init does not work on MAC OS X


biery:
Ran through the README docs for Ubuntu and everything worked well using latest dev build (March 1) and the command: swift build --init.

When running on MAC OS X 10.11.3 I receive an error: 
swift build --init
<unknown>:0: error: no such file or directory: 'build'

Checked the bin folder and noticed the executable was swift-build and this works:
swift-build --init

Command works both ways in Ubuntu, so may be better to go with "swift-build --init" unless there is some background history I am missing.  

This is a small thing, but might frustrate some users if they can't get it to work.

vadimeisenbergibm:
@biery swift build  works for me on MAC OS X 10.11.3. 
Could you please run `swift -v` and paste its output here?

biery:
Apple Swift version 2.1.1 (swiftlang-700.1.101.15 clang-700.1.81)
Target: x86_64-apple-darwin15.3.0
/Applications/Xcode.app/Contents/Developer/usr/bin/lldb "--repl=-target x86_64-apple-darwin15.3.0 -enable-objc-interop -sdk /Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX10.11.sdk -color-diagnostics"

shmuelk:
swift build, only exists as part of Swift 3.0.

You said that you have installed the latest development drop from swift.org. Did you add it to your PATH?

biery:
from .bashrc:

PATH=$PATH:/Applications/IBM/node/bin:/Library/Developer/Toolchains/swift-latest.xctoolchain/usr/bin

Package from swiftlang.org:

swift-DEVELOPMENT-SNAPSHOT-2016-03-01-a-osx.pkg

From swift.org:

https://swift.org/builds/development/xcode/swift-DEVELOPMENT-SNAPSHOT-2016-03-01-a/swift-DEVELOPMENT-SNAPSHOT-2016-03-01-a-osx.pkg

This is the first install I did on this computer, so there were no other versions installed besides this one.    This is on Mac OS X 10.11.3

vadimeisenbergibm:
@biery `/Library/Developer/Toolchains/swift-latest.xctoolchain/usr/bin` should be appended before $PATH in your .bashrc:
```bash
PATH=/Applications/IBM/node/bin:/Library/Developer/Toolchains/swift-latest.xctoolchain/usr/bin:$PATH
```
[...]

PR #282, opened 2016-03-15
Update to README for issue #281

biery:
Unable to run "swift build" on mac os x 10.11.3, must use "swift-build", at least when using latest SWIFT build from swiftlang.org

Ubuntu accepts the command either way.


biery:
issue was with PATH variable pointing to the wrong swift bin.

PR #316, opened 2016-03-23
Bug fix: BodyParser.parse contentType argument is nil.

chrisozenne:
## Description
The capitalized "Content-Type" does not match the lowercase format for storing headers. When using the BodyParser to parse a request, the contentType argument is nil because the "Content-Type" subscript returns nil.

## Motivation and Context
The BodyParser() middleware fails to produce a request body for incoming JSON request with correct "Content-Type" header. The `request.body` property is `nil`.

## How Has This Been Tested?
I made the changes locally with the modification and found that it solved the issue. The `request.body` property was populated with the JSON data I was expecting.

## Checklist:
- [ x] I have submitted a [CLA form](https://github.com/IBM-Swift/CLA)

### DIFF
--- a/Sources/Kitura/bodyParser/BodyParser.swift
+++ b/Sources/Kitura/bodyParser/BodyParser.swift
@@ -45,7 +45,7 @@ public class BodyParser : RouterMiddleware {
     ///
     public func handle(request: RouterRequest, response: RouterResponse, next: () -> Void) {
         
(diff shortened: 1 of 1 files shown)


chrisozenne:
It looks like this may have already been addressed in a commit to Kitura-net earlier today by @shmuelk: https://github.com/IBM-Swift/Kitura-net/commit/61c40ded86caa882e0377e81896e38147993a11d

Issue #86, opened 2016-02-08
Switch to using expectations instead of sleeping in router tests


dfirsht:
<a href="https://github.ibm.com/dfirsht"><img src="https://avatars.github.ibm.com/u/1584?" align="left" width="96" height="96" hspace="10"></img></a> **Issue by [dfirsht](https://github.ibm.com/dfirsht)**
_Tuesday Feb 02, 2016 at 22:09 GMT_
_Originally opened as https://github.ibm.com/ibmswift/Phoenix/issues/86_

----

Currently there is a 10 sec sleep between tests to ensure the server shuts down correctly. It would be preferable to switch this to being notified instead

dfirsht:
This is blocked since XCTest does not provide this functionality. See PR here: https://github.com/apple/swift-corelibs-xctest/pull/43.

modocache:
https://github.com/apple/swift-corelibs-xctest/pull/43 has been merged, so this issue should now be unblocked. Please file JIRA issues for any bugs you encounter, or if you have any suggestions on how to improve the feature! :bow:

dfirsht:
Thanks @modocache, awesome work!

PR #349, opened 2016-04-07
Added expectations to server request test cases

Naithar:
## Description

Added XCTestExpectation to request handler. Changed how server tests is waiting for requests to finish.

## Motivation and Context

IBM-Swift/Kitura#86

## Checklist:

- [] If applicable, I have added tests to cover my changes.

### DIFF
--- a/Tests/Kitura/KituraTest.swift
+++ b/Tests/Kitura/KituraTest.swift
@@ -32,8 +32,6 @@ extension KituraTest {
   //       sleep(10)
     }
 
-
-
     func performServerTest(router: HttpServerDelegate, asyncTasks: (expectation: XCTestExpectation) -> Void...) {
         let server = setupServer(8090, delegate: router)
         let requestQueue = Queue(type: QueueType.SERIAL)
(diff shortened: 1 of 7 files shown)


dfirsht:
You'll need this to pass on Linux as well. Check out this PR for the API for expectations on Linux: https://github.com/apple/swift-corelibs-xctest/pull/43.  You can use `#if os(Linux)` to run system dependent code.

Naithar:
This one fixed it. It's actually sort of strange to see different API for the same library.

dfirsht:
This is great! Everything is working fine. We can merge this is in right now.

modocache:
> It's actually sort of strange to see different API for the same library.

To clarify a bit here: swift-corelibs-xctest provides an API that matches the Swift 3 import rules for Objective-C. You're seeing two different APIs because the version of Swift you use on OS X is a version lower than 3.

Once you've updated to Swift 3 (I assume you're waiting for an official Swift 3 release, which I think is a good idea), you'll be able to use the exact same API across all platforms. In the meantime, sorry for the inconvenience!

I wonder if we can provide both APIs in the meantime, to make it easier for people working with corelibs-xctest now? /cc @mike-ferris-apple @briancroom

modocache:
I created [SR-1215](https://bugs.swift.org/browse/SR-1215) to track this issue. Thanks! Also, feel free to ping me or create a JIRA task for any other feedback you have on swift-corelibs-xctest. 👍

briancroom:
@modocache It looks to me like the issue here isn't actually trying to support Swift 2.2 and 3, but rather that the most recent Swift development snapshot doesn't include the corelibs-xctest changes (from https://github.com/apple/swift-corelibs-xctest/pull/88) required to make its API match the Swift 3 import rules. So this PR needed to use the Swift 2.2 API on Linux and the 3 API on Darwin. Once the next snapshot goes live, things should be unified again. 

That being said, there *may* still be a case for corelibs-xctest exposing an API compatible with the 2.2 naming, but I would want to see a concrete case needing it, and this isn't it, I think.

PR #352, opened 2016-04-11
Added ServerInfo and Server response header

rfdickerson:
Added a class for containing the version of Kitura and also the operating system.

## Description

Added a string that returns "Kitura/0.8.0 (Linux)" if running on Linux.

## Motivation and Context

It is common that in the response headers in HTTP that you include a string representing the Server used like Apache 2, the version, and the platform. 

## Checklist:

---
This change is [Reviewable](https://reviewable.io/reviews/ibm-swift/kitura/352)

### DIFF
--- a/Sources/Kitura/ServerInfo.swift
+++ b/Sources/Kitura/ServerInfo.swift
@@ -1,0 +1,34 @@
+
+/**
+ * Copyright IBM Corporation 2016
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+ 
+public struct ServerInfo {
+   
+   public static let kituraVersion: String = "0.8.0" 
+    
+   #if os(OSX)
+   public static let os = "Mac OS X"
+   #elseif os(Linux)
+   public static let os = "Linux"
+   #else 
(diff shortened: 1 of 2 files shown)


Sources/Kitura/ServerInfo.swift
```diff
@@ -0,0 +1,34 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+ 
+public struct ServerInfo {
+   
+   public static let kituraVersion: String = "0.8.0"
```
ianpartridge: We'll need a good way of updating this as part of the release process.

rfdickerson:
Well, perhaps we should not automatically add the Server tag to the Response header because of the [IETF guideline](http://www.ietf.org/rfc/rfc2068.txt)

Sources/Kitura/ServerInfo.swift
rfdickerson: Agreed, to that effect, I was wondering if the SPM will make build-time variables (like the #if os macro, or the #line and #file macro) but for the current git tag version. Something like #version.  I couldn't find anything like that, sadly.

shmuelk:
We could generate the code at "release" time.

It could be run by a "standard" script that we run that does those sorts of things.....

Sources/Kitura/ServerInfo.swift
ianpartridge: SPM is under active development - we could ask on `swift-build-dev`?

kweinmeister:
Will revisit this when he have a comprehensive plan that includes automation of the version number.

rfdickerson:
Apparently this is a PR that was just merged in 4 days ago:

https://github.com/apple/swift-package-manager/pull/122

Sources/Kitura/ServerInfo.swift
rfdickerson: Thanks! Asking for it on swift-dev got us the answer.

Issue #399, opened 2016-04-20
Move boolean logic to guards in process method of RouterElement


dfirsht:
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/RouterElement.swift#L118-121

dfirsht:
Completed in PR #405

Issue #381, opened 2016-04-19
Remove force unwrapping from Router.swift


dfirsht:
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/Router.swift#L624
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/Router.swift#L643
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/Router.swift#L683

dfirsht:
Completed in PR #405

Issue #377, opened 2016-04-19
Remove default nil assignments for optionals


dfirsht:
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/RouterElement.swift#L35
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/Router.swift#L35
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/Router.swift#L662
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/RouterElement.swift#L30-35
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/RouterElement.swift#L164
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/RouterRequest.swift#L124

dfirsht:
Completed in PR #405

PR #405, opened 2016-04-22
Removed force unwrap, changed if statements to guard, removed default nil assignment

naithar:
## Description
Replaced force unwrapping from ContentType.swift, Router.swift and RouterElement.swift with ```guard let``` statements. 
Moved boolean logic in ```process``` method in RouterElement.swift from ```if``` to ```guard```
Removed default nil assignments in Router.swift, RouterElement.swift and RouterRequest.swift

## Motivation and Context
IBM-Swift/Kitura#381
IBM-Swift/Kitura#399
IBM-Swift/Kitura#377
Removing force unwrapping from ContentType.swift and using ```guard let``` ensures that jsonData value exists and won't crash if it is not conforming to expected type which will happen with ```as!``` and ! operator

## How Has This Been Tested?
Didn't add any new logic that should be tested. All existing tests already cover changes.

## Checklist:


ianpartridge:
Thanks for the patch! @dfirsht - can you review?

dfirsht:
This all looks good to me, I'll merge it in now.

PR #415, opened 2016-04-25
Headers refactor

### LINKED ISSUE #404: Refactor header architecture
In our current architecture we have two arrays, singleHeaders: [String: String] and multiHeaders:[String: [String]]. This is to accommodate the HTTP specification that there may be multiple header fields of the same name.

However this architecture means that we have to access header values through methods: setHeader, setHeaders, getHeader, getHeaders. This can be somewhat unintuitive.

I propose instead, we get rid of singleHeaders have a combined dictionary for headers of type [String: [String]] this gives a more unified approach and would allow us to get rid of getter/settter methods.
Ex. set:
```
header["foo"] = ["bar"]
header["foo"] = ["bar1,"bar2"]
```
Ex. get:
```
let values = header["foo"]
for value in values {
  // do something
}
```
I considered having the array be [String: HeaderValue] where HeaderValue was an enum which could hold either a String or an array of Strings but ultimately this resulted in having to call a switch every time a header was accessed which was clunky.

dfirsht:
## Motivation and Context
Resolves #404. Depends on https://github.com/IBM-Swift/Kitura-net/pull/20 being accepted first.

## Checklist:


Sources/Kitura/RouterRequest.swift
```diff
@@ -32,7 +32,7 @@ public class RouterRequest: SocketReader {
     /// The hostname of the request
     ///
     public var hostname: String {
-        if  let host = headers["host"] {
+        if  let host = headers.get("host")?.first {
```
ianpartridge: Have to say... this looks uglier than it used to.

dfirsht: Yeah it is more verbose but you are getting more from it. The new headers are case-insensitive and can handle multiple headers of the same name. I think in the long run this can help prevent hard to diagnose bugs.

naithar: Wouldn't the use of subscript (```header["host"]```) be better here? Since header is storing and returning values as string array, setting it with ```header["host"] = [value]``` for a single value shouldn't be a problem.

dfirsht: I can add a subscript shorthand so headers["host"] is equivalent to headers.get("host"). It doesn't mean that headers is a full fledged dictionary though.

dfirsht:
This is ready for review again. I have modified the api so headers can be subscripted to receive a string-only value.

Sources/Kitura/Headers.swift
```diff
@@ -0,0 +1,88 @@
...
+        self.headers = headers
+    }
+    
+    public subscript(key: String) -> String? {
+        get {
+            return get(key)
+        }
+        
+        set(newValue) {
+            if let newValue = newValue {
```
ianpartridge: This isn't necessary - `newValue` isn't an optional.

Sources/Kitura/Headers.swift
```diff
@@ -0,0 +1,88 @@
...
+    ///
+    /// Set the header value
+    ///
+    /// - Parameter key: the key
+    /// - Parameter value: the value
+    ///
+    /// - Returns: the value for the key as a list
+    ///
+    public mutating func set(_ key: String, value: String) {
+
```
ianpartridge: Delete this line.

Sources/Kitura/Headers.swift
```diff
@@ -0,0 +1,88 @@
...
+    }
+    
+    ///
+    /// Append values to the header
+    ///
+    /// - Parameter key: the key
+    /// - Parameter value: the value
+    ///
+    public mutating func append(_ key: String, value: String) {
+
```
ianpartridge: Delete this line.

Sources/Kitura/Headers.swift
```diff
@@ -0,0 +1,88 @@
...
+    }
+
+    
+    ///
+    /// Remove the header by key (case insensitive)
+    ///
+    /// - Parameter key: the key
+    ///
+    public mutating func remove(_ key: String) {
+
```
ianpartridge: Delete this line.

Sources/Kitura/Headers.swift
```diff
@@ -0,0 +1,88 @@
...
+    ///
+    /// Remove the header by key (case insensitive)
+    ///
+    /// - Parameter key: the key
+    ///
+    public mutating func remove(_ key: String) {
+        
+        headers.remove(key)
+    }
+}
```
ianpartridge: Add newline?

Sources/Kitura/RouterRequest.swift
```diff
@@ -75,12 +75,7 @@ public class RouterRequest: SocketReader {
     ///
     /// List of HTTP headers with simple String values
     ///
-    public var headers: SimpleHeaders { return serverRequest.headers }
-
-    ///
-    /// List of HTTP headers with String array values
-    ///
-    public var headersAsArrays: ArrayHeaders { return serverRequest.headersAsArrays }
+    public let headers:Headers
```
ianpartridge: Space after colon.

Sources/Kitura/RouterResponse.swift
```diff
@@ -270,85 +273,6 @@ public class RouterResponse {
...
 
     ///
-    /// Gets the header
-    ///
-    /// - Parameter key: the key
-    ///
-    /// - Returns: the value for the key
-    ///
-    public func getHeader(_ key: String) -> String? {
-
```
ianpartridge: Lots of newlines in this file.

Sources/Kitura/Headers.swift
naithar: @ianpartridge actually since return type is optional, ```newValue``` should be optional too. Just checked in playground :)

ianpartridge: You're right - thanks!

ianpartridge:
Overall this looks really nice!  Definitely improvement @dfirsht

Issue #393, opened 2016-04-19
Move ResponseHeadersSetter protocol to its own file


dfirsht:
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/staticFileServer/StaticFileServer.swift#L250-260

dfirsht:
Completed in #424.

Issue #390, opened 2016-04-19
Move ParsedBody to its own file


dfirsht:
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/bodyParser/BodyParser.swift#L140

dfirsht:
Completed in PR #424 .

Issue #386, opened 2016-04-19
Move RouterHandler typealias to its own file


dfirsht:
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/RouterElement.swift#L407

dfirsht:
Completed in PR #424.

Issue #383, opened 2016-04-19
Move RouterMethod enum to its own file


dfirsht:
https://github.com/IBM-Swift/Kitura/tree/develop/Sources/Kitura/RouterElement.swift#L405

dfirsht:
Validated through automatic testing.

PR #424, opened 2016-04-30
Moved code to separate files, removed force unwrap

naithar:
## Description
Moved ```enums```, ```typedefs``` and ```protocols``` to separate files.
Switched optionals force unwrap in multiple files with guard let. 
Switched cookie optional lazy with swift ```lazy``` property in RouterRequest.swift

## Motivation and Context
IBM-Swift/Kitura#390
IBM-Swift/Kitura#386
IBM-Swift/Kitura#383
IBM-Swift/Kitura#382
IBM-Swift/Kitura#393
Usage of guard let is more save than force unwrap. Usage of Switf ```lazy``` property ensures variable is initialized single time and removes the need to unwrap it with ```guard```/```if```.

## How Has This Been Tested?
All existing tests are already covering this changes. 

## Checklist:


Sources/Kitura/Router.swift
```diff
@@ -726,8 +733,11 @@ extension Router : HttpServerDelegate {
             sendResourceIfExisting(routeResp, resource: "index.html")
         } else {
             do {
-                try routeResp.status(HttpStatusCode.NOT_FOUND).send("Cannot \(String(routeReq.method).uppercased()) \(routeReq.parsedUrl.path ?? "").").end()
-            } catch {}
+                let message = "Cannot \(String(routeReq.method).uppercased()) \(routeReq.parsedUrl.path ?? "")."
+                try routeResp.status(HttpStatusCode.NOT_FOUND).send(message).end()
```
ianpartridge: This will conflict against https://github.com/IBM-Swift/Kitura-net/pull/22 so depending on merge order we will need to fix up.

ianpartridge:
Looks good to me!  Once everyone is happy, we should probably squash the commits before merge.

Sources/Kitura/Router.swift
naithar: If this pull request will follow IBM-Swift/Kitura-net#22, which will probably happen, I'll fix it :)

Sources/Kitura/RouterRequest.swift
```diff
@@ -95,16 +95,15 @@ public class RouterRequest: SocketReader {
     //
     // Parsed Cookies, used to do a lazy parsing of the appropriate headers
     //
-    private var _cookies: Cookies?
+    private lazy var _cookies: Cookies = {
```
naithar: This closure could probably use ```unowned self``` or ```weak self``` so there won't be a retain cycle. But I'm not sure retain cycle would happen at this point.

dfirsht:
Hey @naithar looks like I accidentally duplicated some of your work. If you can resolve the merge conflicts and squash the commits, I can accept this right away.

naithar:
@dfirsht merged changes. Done with wrong account but hopefully it won't be a problem :)

Issue #311, opened 2016-03-22
Add link method for RouterResponse


dfirsht:
See Express link method for a model: http://expressjs.com/en/4x/api.html#res.links.

kweinmeister:
Moved to In Progress because of PR. If it doesn't pass checks or not desired in 0.14, feel free to move back to Product Backlog.

dfirsht:
Completed with PR #440.

PR #440, opened 2016-05-05
Link method for RouterResponse

naithar:
## Description

Added ```link``` method for RouterResponse class.

## Motivation and Context

IBM-Swift/Kitura#311

## How Has This Been Tested?

Added a test in TestResponse.swift file that checks received Link header value for both single and multiple links.

## Checklist:

### DIFF
--- a/Sources/Kitura/RouterResponse.swift
+++ b/Sources/Kitura/RouterResponse.swift
@@ -501,17 +501,55 @@ public class RouterResponse {
     }
 
     ///
-    /// Sets Link HTTP header
+    /// Adds a link with specified parameters to Link HTTP header
     ///
-    /// - Parameter links: a dictionary that maps link rel parameter to actual link value
+    /// - Parameter link: link value
+    /// - Parameter rel:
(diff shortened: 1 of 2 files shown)


dfirsht:
Hey @naithar, sorry for the delay, we have been discussing this PR internally. The issue is that if you look at the spec for Link (https://tools.ietf.org/html/rfc5988) you can see that it can take in a lot more parameters than ref. After looking at the implementation of Link in JAX, and seeing some PRs for changing Link functionality in Express, we decided that we should support all options. 

Therefore we propose that the Link function will instead take in one link at a time, and have all possible values for link be accepted as nil default optionals.

The new signature will be something like `func link (_ link: String, rel: String? = nil, anchor: String? = nil, title: String? = nil, type: String? = nil, media: String? = nil)`

naithar:
@dfirsht, thanks for the info. I'll change the implementation then.

naithar:
@dfirsht please review

dfirsht:
Looks great! Thanks @naithar

PR #460, opened 2016-05-10
Fixed import statement for 'Kitura_TemplateEngine'

### LINKED ISSUE #459: No such module 'KituraTemplateEngine'
Getting this error in `Router.swift` on version 0.13.1. This error only appears in Xcode. When using `swift build` from the command line, the build is successful.

AnthonyMDev:
Closes #459

### DIFF
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -18,7 +18,7 @@ import KituraNet
 import KituraSys
 import LoggerAPI
 import Foundation
-import KituraTemplateEngine
+import Kitura_TemplateEngine
 
 // MARK Router
 


AnthonyMDev:
So, after looking at this a little further, by changing the import statement to `import Kitura_TemplateEngine`, the build error goes away in Xcode, but running `swift build` from the command line gives the following error:

`/MyApp/Packages/Kitura-0.13.1/Sources/Kitura/Router.swift:21:8: error: no such module 'Kitura_TemplateEngine'
import Kitura_TemplateEngine`

If I go back to `import KituraTemplateEngine`, it builds fine from the command line, but the Xcode build error comes back.

kweinmeister:
We'll need a fix that works both on the CLI and Xcode.

I haven't researched this thoroughly yet, but I wonder if the empty Package.swift in Kitura-TemplateEngine is triggering an edge case error with `swift build -X`. We have other modules with the Kitura-* naming scheme, but they have a "name" attribute in Package.swift.

AnthonyMDev:
It might be best that we add a name to the package that does not include the hyphen for importing?

shmuelk:
Thanks for your interest in contributing to Kitura.

I have corrected the original issue in a more proper fashion, by adding to Kitura-TemplatingEngine a "proper target" (KituraTemplatingEngine). I simply moved the one source code file to an added directory under Sources.

PR #471, opened 2016-05-15
Fixed KituraTemplateEngine import error in Xcode

alexaubry:
## Description
Because of a dash in the template engine's name, Xcode automatically changes `-` to `_`. Using the `Xcode` compiler macro, that is present in all projects generated using the SPM command line tool, to reflect this change and provide the correct module name in Xcode fixes the compiling issue, without breaking existing source or impacting command line builds.

## Motivation and Context
When using Xcode to build a Kitura-based project, compiling fails because the module `KituraTemplateEngine` can't be found. Not only does this error affect the building process and the dependency tree, but it also leads to crash/hangs and _Xcode internal errors._

## How Has This Been Tested?
As mentioned above, this patch does not impact the code itself, performing functional tests was not relevant. However I did perform building tests :

- Building with `swift build` : succeeded
- Building inside Xcode : succeeds

Building on Linux should pass too, wasn't able to test it though.

## Checklist:

- [X] I have submitted a [CLA form](https://github.com/IBM-Swift/CLA)
- [ ] If applicable, I have updated the documentation accordingly. (not applicable)
- [ ] If applicable, I have added tests to cover my changes. (not applicable)

### DIFF
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -18,7 +18,12 @@ import KituraNet
 import KituraSys
 import LoggerAPI
 import Foundation
+
+#if Xcode
+import Kitura_TemplateEngine
+#else
 import KituraTemplateEngine
+#endif
 
 // MARK Router
 


shmuelk:
This was fixed in issue #459. Make sure you are running with KIturaTemplate 0.13.x

Sources/Kitura/Router.swift
```diff
@@ -18,7 +18,12 @@ import KituraNet
 import KituraSys
 import LoggerAPI
 import Foundation
+
+#if Xcode
+import Kitura_TemplateEngine
+#else
 import KituraTemplateEngine
+#endif
```
aciidb0mb3r: This doesn't seem correct, once https://github.com/apple/swift-package-manager/pull/337 is merged, the module name will be `Kitura_TemplateEngine` for both SwiftPM and Xcode.

Some context: 
Xcode converts all invalid characters in a module name to `_` where as SwiftPM was removing them instead. That PR will match the module name with Xcode's.

shmuelk: @aciidb0mb3r What the fix diff was add a more explicit target to the directory structure.

The original repository had source files directly under the Sources directory. In such cases SPM seems to set the module name to the value of the name attribute in the Package.swift file. As the name in the Package.swift file is Kitura-TemplateEngine, the issues you bring up were relevant.

The fix simply added a directory KituraTemplateEngine under the Sources directory and moved the source files under there. That causes SPM to create a module with the name KituraTemplateEngine, whether it is building by itself or generating an XCode project.

shmuelk:
This was fixed as mentioned in the comment above.

Thank you very much for your interest in Kitura and for taking the time to try and fix a bug in Kitura.

Issue #483, opened 2016-05-18
Implement multipart form request parser


dfirsht:

dsperling:
@dfirsht - Nice work!  I know this is still work in progress, but in my testing I have seen one issue I wanted to pass along.

If the parts of a POST include a Content-Length header field, the line containing the Content-Length is included in the part data.  I the example below, there are 2 Parts:

The "requestid" part is .text and includes:
```
Content-Length: 36

180325d6-f3de-42e3-a0c6-56e72800c02c
```
The "file" part is returned as .raw instead of .json and also includes the Content-Type in the returned NSData.
```
POST /example HTTP/1.1
Content-Type: multipart/form-data;boundary=BOUNDARY-EXAMPLE

--BOUNDARY-EXAMPLE
Content-Disposition: form-data; name="requestid"
Content-Type: text/plain;charset=ISO-8859-1
Content-Length: 36

180325d6-f3de-42e3-a0c6-56e72800c02c
--BOUNDARY-EXAMPLE
Content-Disposition: form-data; name="file"; filename="message.json"
Content-Type: application/json
Content-Length: 44

{"result": 1, "date": "2016-05-18 12:02:03"}

--BOUNDARY-EXAMPLE--

```
I don't have a suggested fix, but will be looking into it today.

dfirsht:
Thanks for the feedback, @dsperling. I implemented this parser following the protocol as described here: http://www.faqs.org/rfcs/rfc2388.html. It is my understanding that `Content-Length` can not be specified in the parts themselves and instead the total `Content-Length` should be set in the request headers. Can you point me to some documentation that shows otherwise?

dsperling:
I found this issue when integrating a web service API from a partner company.  It is possible that the partner's web service implementation is out of spec.

I will take a look for documentation on the headers.  I guess this issue would be a problem if a developer inserted any arbitrary header as well.

dsperling:
Content-Type indeed should not be included, nor should be evaluated.  However, the specification does allow for custom headers.  Here are the relevant sections of RFC 238 and 2046:
```
RFC 2388

3. Definition of multipart/form-data

   The media-type multipart/form-data follows the rules of all multipart
   MIME data streams as outlined in [RFC 2046]. 


RFC 2046

5.1.  Multipart Media Type

   ...
   After its boundary delimiter line, each body part then consists of a
   header area, a blank line, and a body area.  ...

   ...  A body part that starts with a
[...]

PR #502, opened 2016-06-01
Improvement for issue #483 - ignore multipart unsupported headers

dsperling:
## Description
Improvements to the feature added in issue #483

## Motivation and Context
In testing the fix for issue #483, it was found that the implementation was not ignoring extension headers. 

## How Has This Been Tested?
Tested with unit tests and web services that include non-standard headers in multipart forms.

## Checklist:

- [ x ] I have submitted a [CLA form](https://github.com/IBM-Swift/CLA)
- [ x ] If applicable, I have added tests to cover my changes.


Sources/Kitura/bodyParser/BodyParser.swift
```diff
@@ -215,12 +215,15 @@ public class BodyParser: RouterMiddleware {
                     // end boundary found, end of parsing
                     return .multipart(parts)
                 }
-                state = .body
+                state = .header
+            }
+
```
dfirsht: If the line has boundary data it should not fall to the switch, instead we should add a `continue`.

Sources/Kitura/bodyParser/BodyParser.swift
```diff
@@ -236,22 +239,22 @@ public class BodyParser: RouterMiddleware {
...
-                        if partData.length > 0 {
-                            // data is multiline, add linebreaks back in
-                            partData.append(newLineData)
-                        }
-                        partData.append(bodyLine)
+                    else if bodyLineAsString.isEmpty {
+                        // empty line signals the end of headers
+                        state = .body
                     }
+                    // skip custom headers
```
dfirsht: Custom headers should still be retained and passed along, we should either expand the Header type to include .other(String) where the associated value is the unsupported header.

Tests/Kitura/TestResponse.swift
```diff
@@ -150,11 +150,41 @@ class TestResponse : XCTestCase {
                     "-----------------------------9051914041544843365972754266\r\n" +
                     "Content-Disposition: form-data; name=\"file1\"; filename=\"a.txt\"\r\n" +
                     "Content-Type: text/plain\r\n\r\n" +
-                    "Content of a.txt.\r\n\r\n" +
+                    "Content of a.txt.\r\n" +
```
dfirsht: I do not believe there is any rule stating that you cannot have double cartage returns at the end like this, I do not see why it needs to be eliminated.

Sources/Kitura/bodyParser/BodyParser.swift
dsperling: Yes, I introduced this bug.

Sources/Kitura/bodyParser/BodyParser.swift
dsperling: Good idea.  That had not occurred to me.  We should add this as a test case as well.

Tests/Kitura/TestResponse.swift
dsperling: Yes, multiple empty lines are supported, but then they would end up in the data reflected back by the test server.  So if the \r\n were added back to "Content of a.txt\r\n", it would then appear in the response:

`XCTAssertEqual(body!, "text text(\"text default\") file1 text(\"Content of a.txt.\r\n\") file2 text(\"<!DOCTYPE html><title>Content of a.html.</title>\") ")
`

I removed it from the POST to keep the test case the same.

dsperling:
I found another issue with the boundary variable.  If the Content-Type has a trailing parameter, then the boundary includes it.  The result is zero parts parsed.

`Content-Type: 	multipart/form-data; boundary=ZZZY70gRGgDPOiChzXcmW3psiU7HlnC; charset=US-ASCII`

I will work on a PR to fix this.

dsperling:
Added the "continue" and fixed the trailing parameter issue on the boundary.

dfirsht:
Hey @dsperling, thanks for the fix. Can you add in the other suggestion I made about the `other(string)` enum?

dsperling:
@dfirsht - the current design would only allow one custom header to be added via .other.  It would also be inconstant with the headers design from RouterRequest.  Should we integrate the design from Headers.swift or defer this feature?

dfirsht:
Well since it's an array you could have any number of custom headers. However, you bring up a good point that the headers design is inconsistent so perhaps it would be nicer to just use the Headers class. Since this PR has been open for so long, I think we should merge it in now and deal with the custom headers as a separate issue. @dsperling can you fix the merge conflicts so we can merge?

dsperling:
There is a change in MultiPartBodyParser that did not get merged correctly.  I will get this fixed tonight during Game 6.

dsperling:
@dfirsht - This update ported the original fix to MultiPartBodyParser.swift with minimal code changes.

I am also working on a refactor to improve MultiPartBodyParser.swift.  These changes should meet the performance goals of what you and @shmuelk were discussing in the thread of issue #483.

I will be ready to propose this improvement tonight.  Should I add it here to this PR or create a new one?

dfirsht:
@dsperling Awesome! This looks really good. I think that we should merge this PR in now and you can make a new PR for the performance improvements afterwards.

Issue #310, opened 2016-03-22
Add JSONP support for RouterResponse


dfirsht:

tunniclm:
I accidentally made the changes based on master, but this is what they look like at the moment:
https://github.com/IBM-Swift/Kitura/compare/master...tunniclm:add_jsonp

tunniclm:
Rebased my changes on top of develop:
https://github.com/IBM-Swift/Kitura/compare/develop...tunniclm:add_jsonp_on_develop

ianpartridge:
Merged in 0430d9d356e9ed8c339c4a6d6b1e4d47da481e56

PR #533, opened 2016-06-14
Add JSONP support

tunniclm:
Add jsonp support to RouterResponse send()

## Description
Added a new overload to `send(jsonp:callbackParam:="callback")` which writes a JSONP response.
Throws a JSONPError if no valid parameter is provided by the request that matches the JSONP callback param name.

## Motivation and Context
Issue #310

## How Has This Been Tested?
The changes include a test route and 3 test cases that check the generated response.
I ran these tests on Linux using the Kitura vagrant image, with Swift snapshot 2016-05-09.
There should probably be some testing in the browser before this is landed.

## Checklist:

### DIFF
--- a/Sources/Kitura/JSONPError.swift
+++ b/Sources/Kitura/JSONPError.swift
@@ -16,12 +16,12 @@
 
 import Foundation
 
-enum JSONPError: ErrorProtocol {
+public enum JSONPError: ErrorProtocol {
     case invalidCallbackName(name: String?)
 }
 
 extension JSONPError: CustomStringConvertible {
-    var description: String {
+    public var description: String {
         switch self {
         case .invalidCallbackName(let name):
             return "Invalid callback name \(name)"
--- a/Sources/Kitura/JSONPError.swift
+++ b/Sources/Kitura/JSONPError.swift
@@ -1,0 +1,30 @@
+/**
+ * Copyright IBM Corporation 2016
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ **/
+
+import Foundation
+
+enum JSONPError: ErrorProtocol {
+    case invalidCallbackName(name: String?)
+}
+
+extension JSONPError: CustomStringConvertible {
+    var description: String {
+        switch self {
+        case .invalidCallbackName(let name):
+            return "Invalid callback name \(name)"
+        }
+    }
+}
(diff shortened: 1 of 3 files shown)


dfirsht:
@tunniclm looks good, just please fix the compilation errors from using the latest the codebase.

tunniclm:
Updated.

tunniclm:
So the test I added fails on OSX because it checks for a specific JSON serialized string value to be returned in the response. However, it looks like NSJSONSerialization does slightly different "pretty" serialization on OSX and Linux (the OSX implementation puts a space before the `:` character in maps and whereas the Linux implementation doesn't).

The doc for NSJSONSerialization is not specific about the exact spacing requirements, so both implementations match the "spec". Nevertheless, I think it would be good for them to match exactly, so I'll raise a PR against swift-corelibs-foundation for that.

This does raise the question of whether or not we want to `send(json:)` and `send(jsonp:)` to produce pretty JSON by default, and/or whether we want to make it a configurable option (like Express, which has an application setting "json spaces").

tunniclm:
Should add a test to check invalid JS characters get correctly encoded
Should add a test that checks changing `callbackParameter` works
Should manually test out in a basic browser app that uses JSONP

tunniclm:
Pushed a new commit with tests for the callback parameter name and the character encoding.

tunniclm:
I ran some manual tests with a browser and it seems to be working correctly.

Here is the test server code:
```swift
import Kitura                                                                                                                                                                
import SwiftyJSON

let router = Router()

router.get("/test") { req, res, next in
    do {
        try res.send(jsonp: JSON([ "hello": "world}" ]))
    } catch (JSONPError.invalidCallbackName) {
        res.status(.badRequest)
    }
    next()
}

Kitura.addHTTPServer(onPort: 8090, with: router)
Kitura.run()
```

And here is the client html page:

```html
<!DOCTYPE html>
<div id="message">Loading...</div>
<script>
function process(json) {
    document.getElementById("message").innerHTML = JSON.stringify(json)
}
</script>

<script src="http://localhost:8090/test?callback=process"></script>
```

The expected output is for the page to show `{ "hello": "world" }`.

I also ran a couple of extra tests:
* modified server to send JSON containing U+2028 and checked it came out correctly in the data the client received
* modified html client to request from `http://localhost:8090/test?callback__wrongname=process` to ensure HTTP 400 Bad Request was returned (server expects a parameter called `callback` which is missing in this request)
* modified html client to request from `http://localhost:8090/test?callback=proc!ess` to ensure HTTP 400 Bad Request was returned (server expects the value of the callback parameter to only contain certain valid characters -- consistent with describing a valid javascript identifier -- this request passes `pro!cess` which is not valid as `!` is not one of the allowed character)

tunniclm:
@shmuelk @dfirsht : I think this is ready to merge provided the latest code changes look good.

Highlights of the most recent changes:

* After much fiddling with inconsistencies in JSON handling between Linux/macOS, I just settled on conditionally compiling the code with `#if`
* I changed  `JSONPError` to be a public enum since it gets thrown up to user code (this didn't show up as a problem in the tests because they are local to Kitura, but **did** give an error from a separate server project.)

tunniclm:
Another quick update to the code as I spotted I was using mime type `application/json` when it should have been `application/javascript`. Fixed that and added some asserts to the tests to test for it as well.

tunniclm:
Updated the doc comment and @ianpartridge reviewed it. I don't think there are any other documentation updates required so I'm going to check off that task as complete.

dfirsht:
Hey we are actually moving all development to be PRs against master. Please reopen your PR against the master branch.

PR #637, opened 2016-07-26
Issue 254

vadimeisenbergibm:
@shmuelk Please review and merge. Since Ray Wenderlich Swift Style guide does not have license, it is better not to copy it.


shmuelk:
As we agreed.

PR #659, opened 2016-07-29
Improved the directions for installing libDispatch

rfdickerson:
## Description

The instructions for installing libDispatch was good in the sense it was simple for copying and pasting into the Terminal. However, with each toolchain path being different, it might be easier to define the path to the toolchain outside of the configuration command.

Idea comes from lots of experience simplifying the process for myself and others. @chiahuang 

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,8 +81,13 @@ Kitura is tested on Ubuntu 14.04 LTS and Ubuntu 15.10.
  Follow the instructions provided on that page. After installing it (i.e. uncompressing the tar file), make sure you update your PATH environment variable so that it includes the extracted tools: `export PATH=/<path to uncompress tar contents>/usr/bin:$PATH`. To update the PATH env variable, you can update your [.bashrc file](http://www.joshstaiger.org/archives/2005/07/bash_profile_vs.html).
 
 3. Clone, build and install the libdispatch library.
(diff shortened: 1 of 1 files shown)


rfdickerson:
Haha, how can changing the README cause the build to fail?

youming-lin:
@rfdickerson I restarted the build and it passed this time. Guessing it was a timing issue with one of the tests.

ianpartridge:
I like the general idea, but unfortunately fenced code blocks in Markdown don't wrap, which means the long line disappears off the right of the page. We need people to be able to read it :(

youming-lin:
I've been thinking: can we upload a script that executes those commands and tell users to run this script instead of having users copy and then execute those commands? I think it'd be more user friendly this way.

rfdickerson:
Fixed!

akrabat:
> I've been thinking: can we upload a script that executes those commands and tell users to run this script instead of having users copy and then execute those commands? I think it'd be more user friendly this way.

I thought the plan was that libdispatch would be part of the Swift distribution by 3.0?

ianpartridge:
> I thought the plan was that libdispatch would be part of the Swift distribution by 3.0?

Yes it is, see https://lists.swift.org/pipermail/swift-corelibs-dev/Week-of-Mon-20160725/000812.html

simplepanda:
Totally off topic but nice to see (in the link @ianpartridge posted above) that URLSession is finally getting implemented. That'll save enormous hassle for me (and I'm guessing a lot of people developing on OS X and deploying on Linux).

PR #825, opened 2016-11-01
Easier access to ParsedBody elements

shmuelk:
## Description
Eliminate the need to use a switch statement to gain access to the information in the ParsedBody enum.

## Motivation and Context
Consumability of Kitura.

## How Has This Been Tested?
Some of Kitura's tests have been modified to use the new APIs.

## Checklist:

### DIFF
--- a/Sources/Kitura/bodyParser/ParsedBody.swift
+++ b/Sources/Kitura/bodyParser/ParsedBody.swift
@@ -50,14 +50,12 @@ public indirect enum ParsedBody {
     /// - Returns: The parsed body as a JSON object, or nil if the body wasn't in
     ///           JSON format.
     public var asJSON: JSON? {
-        let result: JSON?
         switch(self) {
         case .json(let body):
-            result = body
+            return body
         default:
-            result = nil
+            return nil
         }
-        return result
     }
     
     /// Extract a "multipart" body from the `ParsedBody` enum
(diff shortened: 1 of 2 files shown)


Sources/Kitura/bodyParser/ParsedBody.swift
```diff
@@ -44,4 +44,63 @@ public indirect enum ParsedBody {
     /// If the content type was "multipart/form-data" this associated value will
     /// contain an array of parts of multi-part respresentation of the body.
     case multipart([Part])
+    
+    /// Extract a "JSON" body from the `ParsedBody` enum
+    ///
+    /// - Returns: The parsed body as a JSON object, or nil if the body wasn't in
+    ///           JSON format.
+    public var asJSON: JSON? {
```
ianpartridge: I would have done:

```swift
public var asJSON: JSON? {
    switch(self) {
    case .json(let body):
        return body
    default:
         return nil
    }
}
```

but no big deal :)

Sources/Kitura/bodyParser/ParsedBody.swift
```diff
@@ -44,4 +44,63 @@ public indirect enum ParsedBody {
...
+        let result: JSON?
+        switch(self) {
```
ianpartridge: Are parentheses required here?

Issue #818, opened 2016-10-27
Support mergeParams


youming-lin:
Node Express supports `mergeParams` to enable subrouters to access matched params defined in the parent router.

http://expressjs.com/en/api.html#express.router

This is functionality that we should support in order to maintain parity with Express.

youming-lin:
Functionality has been added via #838.

PR #838, opened 2016-11-04
Add mergeParams support and more unit tests https://github.com/IBM-Sw…

youming-lin:
Add support for `mergeParams` in `Router`

## Motivation and Context

https://github.com/IBM-Swift/Kitura/issues/818

## How Has This Been Tested?

Added unit tests to validate behavior of `mergeParams`.

## Checklist:

…ift/Kitura/issues/818

### DIFF
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -149,10 +149,11 @@ public class Router {
     /// path and handles all of the mappings of paths below that.
     ///
     /// - Parameter route: The path to bind the sub router to.
-    ///
+    /// - parameter mergeParams: Specify if this router should have access to path parameters
+    /// matched in its parent router. Defaults to `false`.
     /// - Returns: The created sub router.
-    public func route(_ route: String) -> Router {
-        let subrouter = Router()
+    public func route(_ route: String, mergeParams: Bool = false) -> Router {
+        let subrouter = Router(mergeParams: mergeParams)
         self.all(route, middleware: subrouter)
         return subrouter
     }
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -51,13 +51,13 @@ public class Router {
     fileprivate let fileResourceServer = FileResourceServer()
 
     /// Flag to enable/disable access to parent router's params
-    private let mergeParams: Bool
+    private let mergeParameters: Bool
 
     /// Initialize a `Router` instance
-    /// - parameter mergeParams: Specify if this router should have access to path parameters
+    /// - parameter mergeParameters: Specify if this router should have access to path parameters
     /// matched in its parent router. Defaults to `false`.
-    public init(mergeParams: Bool = false) {
-        self.mergeParams = mergeParams
+    public init(mergeParameters: Bool = false) {
+        self.mergeParameters = mergeParameters
 
         Log.verbose("Router initialized")
     }
@@ -66,7 +66,7 @@ public class Router {
         elements.append(RouterElement(method: method,
                                       pattern: pattern,
                                       handler: handler,
-                                      mergeParams: mergeParams))
+                                      mergeParameters: mergeParameters))
         return self
     }
 
@@ -75,7 +75,7 @@ public class Router {
                                       pattern: pattern,
                                       middleware: middleware,
                                       allowPartialMatch: allowPartialMatch,
-                                      mergeParams: mergeParams))
+                                      mergeParameters: mergeParameters))
         return self
     }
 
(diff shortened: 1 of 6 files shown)


youming-lin:
@shmuelk Can you review this?

Tests/KituraTests/TestRouteRegex.swift
```diff
@@ -90,4 +115,122 @@ class TestRouteRegex: XCTestCase {
...
+                    XCTFail()
+                }
+
+                expectation.fulfill()
+            })
+        })
+
+        router = Router()
+
+        router.all("/*", middleware: TestRouteRegex.subrouter)
```
shmuelk: While what you wrote to add a sub-router works. Why didn't you use the Router.route to create and mount a sub-router. The route function should also get an extra parameter to pass in the mergeParameters flag.

Sources/Kitura/RouterElement.swift
```diff
@@ -124,7 +134,8 @@ class RouterElement {
     /// - Parameter match: the regular expression result
     /// - Parameter request:
     private func setParameters(forRequest request: RouterRequest, fromUrlPath urlPath: NSString, match: TextChekingResultType) {
-        var parameters = [String:String]()
+        var parameters = mergeParams ? request.parameters : [:]
```
shmuelk: Once in a router that is having the parameters merged, each and every subsequent RouteHAndler or middleware will get the cumulative set of the parameters from the top level router and every path that matched. The original design was that each RouteHandler or middleware gets its own parameters and no-one else's. I understand that here we adding in the parameters from the parent route, but are we adding in all of the previous routes as well.
Some examples might help.
```swift
  let router = Router()
  router.get("/path/:p1/:p2", handler: handler1)
  router.get("/path/:p1/testing", handler: handler2)
``` 
In this case, for the same URL path, both handler1 and handler2 will be invoked. When handler1 is invoked there will be two elements in request.parameters, p1 and p2. When handler2 is invoked there will be one element in requests.parameters, p1.

```swift
  let router = Router()
  let subRouter = router.route("/path/:p1", mergeParameters: true)
  subRouter.get("/:p2", handler: handler1)
  subRouter.get("/testing", handler: handler2)
``` 
In this case, for the same URL path, both handler1 and handler2 will be invoked. When handler1 is invoked there will be two elements in request.parameters, p1 and p2. But now when handler2 is invoked there will **also** be **two* elements in requests.parameters, p1 and p2, **instead of** just p1.

Tests/KituraTests/TestRouteRegex.swift
youming-lin: I updated `TestSubrouter` to use `Router.route` but left it as it is in `TestRouteRegex`. I'm reusing the same subrouter on different mount paths so I didn't use `Router.route` here.

youming-lin:
@shmuelk Can you review this again?

Sources/Kitura/Router.swift
```diff
@@ -149,10 +149,11 @@ public class Router {
     /// path and handles all of the mappings of paths below that.
     ///
     /// - Parameter route: The path to bind the sub router to.
-    ///
+    /// - parameter mergeParams: Specify if this router should have access to path parameters
+    /// matched in its parent router. Defaults to `false`.
     /// - Returns: The created sub router.
-    public func route(_ route: String) -> Router {
-        let subrouter = Router()
+    public func route(_ route: String, mergeParams: Bool = false) -> Router {
```
shmuelk: Please make the new parameter ```mergeParameters```.

Sources/Kitura/Router.swift
```diff
@@ -149,10 +149,11 @@ public class Router {
...
+        let subrouter = Router(mergeParams: mergeParams)
```
shmuelk: Please change this to read ```let subrouter = Router(mergeParameters: mergeParameters)```

I'm sorry I didn't catch this is in the first review.

Tests/KituraTests/TestSubrouter.swift
```diff
@@ -156,33 +156,20 @@ class TestSubrouter: XCTestCase {
...
-        subRouter2.get("/sub1/:sub1", middleware: subsubRouter1)
-
         let router = Router()
+        let subsubRouter1 = router.route("/root1/:root1").route("/sub1/:sub1", mergeParams: true)
 
-        router.get("/root1/:root1", middleware: subRouter1)
-        router.get("/root2/:root2", middleware: subRouter2)
+        subsubRouter1.all("/subsub1/:subsub1", handler: handler)
+        subsubRouter1.all("/subsub2/:subsub2", handler: simpleHandler)
+        subsubRouter1.all("/subsub2/passthrough", handler: handler)
```
shmuelk: I like the tests. 👍

PR #480, opened 2016-05-18
Parameter method for Router

naithar:
## Description
Implementer ```parameter``` method for ```Router``` class that can be used to specify handlers for specific request parameters.

## Motivation and Context
IBM-Swift/Kitura#295

## How Has This Been Tested?
Added test ```testRouterParameter``` that cover this changes.

## Checklist:

- [x] I have submitted a [CLA form](https://github.com/IBM-Swift/CLA)
- [ ] If applicable, I have updated the documentation accordingly.
- [x] If applicable, I have added tests to cover my changes.

Issue #295, opened 2016-03-22
Add functionality to specify param specific middleware


dfirsht:
Example from Express:
router.param('user', function(req, res, next, id) {
  console.Log(':user is present in route path');
  next();
});

naithar:
I have started implementing this feature and it seems to be working without any problems on Mac, but Linux tests are failing at this moment.
https://github.com/naithar/Kitura/blob/develop/Sources/Kitura/Router.swift#L610-622
https://github.com/naithar/Kitura/blob/develop/Sources/Kitura/RouterParameter.swift
https://github.com/naithar/Kitura/blob/develop/Sources/Kitura/RouterMiddlewareWalker.swift

shmuelk:
We will review and hopefully merge the above PR early next week in our next sprint

naithar:
@shmuelk I tried to reimplement ```parameter``` method again with express.js in mind, but currently I'm not sure it is alright to make a PR. 
The problem is actually that I'm not sure that parameter handlers should be stored in separate list and be passed down to element looper and element's ```process``` method as i currently did. If there is no problem with such implementation, I'll clean the code and make a PR for review.
https://github.com/naithar/Kitura/blob/develop/Sources/Kitura/Router.swift#L717-721
https://github.com/naithar/Kitura/blob/develop/Sources/Kitura/RouterElementWalker.swift#L55-58
https://github.com/naithar/Kitura/blob/develop/Sources/Kitura/RouterElement.swift#L134-139

PR #849, opened 2016-11-14
Router .parameter method implementation

naithar:
Another try to implement a `.parameter` method for `Router` to handle parsed parameters.
Previous implementation: IBM-Swift/Kitura#480

## Motivation and Context
IBM-Swift/Kitura#295

## How Has This Been Tested?
Added test to `TestRequests.swift` file

## Checklist:

### DIFF
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -161,19 +161,34 @@ public class Router {
         return subrouter
     }
 
+    /// Handle an URL parameter
     ///
+    /// - Parameter name: A single parameter name to be handled
+    /// - Parameter handler: A comma delimited set of `RouterParameterHandler`s that will be
+    ///                     invoked when request parses a parameter with specified name.
+    /// - Returns: Current router instance
     @discardableResult
     public func parameter(_ name: String, handler: @escaping RouterParameterHandler...) -> Router {
         return self.parameter([name], handlers: handler)
     }
 
+    /// Handle an URL parameter
     ///
+    /// - Parameter names: The array of parameter names that will be used to invoke handlers
+    /// - Parameter handler: A comma delimited set of `RouterParameterHandler`s that will be
+    ///                     invoked when request parses a parameter with specified name.
+    /// - Returns: Current router instance
     @discardableResult
     public func parameter(_ names: [String], handler: @escaping RouterParameterHandler...) -> Router {
         return self.parameter(names, handlers: handler)
     }
 
+    /// Handle an URL parameter
     ///
+    /// - Parameter names: The array of parameter names that will be used to invoke handlers
+    /// - Parameter handlers: The array of `RouterParameterHandler`s that will be
+    ///                     invoked when request parses a parameter with specified name.
+    /// - Returns: Current router instance
     @discardableResult
     public func parameter(_ names: [String], handlers: [RouterParameterHandler]) -> Router {
         for name in names {
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -53,7 +53,9 @@ public class Router {
     /// Flag to enable/disable access to parent router's params
     private let mergeParameters: Bool
 
-    ///
+    /// Collection of `RouterParameterHandler` for specified parameter name
+    /// that will be passed to `RouterElementWalker` when server receives client request
+    /// and used to handle request's url parameters.
     fileprivate var parameterHandlers = [String : [RouterParameterHandler]]()
 
     /// Initialize a `Router` instance
@@ -161,7 +163,10 @@ public class Router {
         return subrouter
     }
 
-    /// Handle an URL parameter
+    // MARK: Parameter handling
+
+    /// Setup a  handler for specific name of request parameters.
+    /// This can make it easier to handle values of provided parameter name.
     ///
     /// - Parameter name: A single parameter name to be handled
     /// - Parameter handler: A comma delimited set of `RouterParameterHandler`s that will be
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -177,7 +177,8 @@ public class Router {
         return self.parameter([name], handlers: handler)
     }
 
-    /// Handle an URL parameter
+    /// Setup a  handler for specific name of request parameters.
+    /// This can make it easier to handle values of provided parameter name.
     ///
     /// - Parameter names: The array of parameter names that will be used to invoke handlers
     /// - Parameter handler: A comma delimited set of `RouterParameterHandler`s that will be
@@ -188,7 +189,8 @@ public class Router {
         return self.parameter(names, handlers: handler)
     }
 
-    /// Handle an URL parameter
+    /// Setup a  handler for specific name of request parameters.
+    /// This can make it easier to handle values of provided parameter name.
     ///
     /// - Parameter names: The array of parameter names that will be used to invoke handlers
     /// - Parameter handlers: The array of `RouterParameterHandler`s that will be
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -53,6 +53,9 @@ public class Router {
     /// Flag to enable/disable access to parent router's params
     private let mergeParameters: Bool
 
+    ///
+    fileprivate var parameterHandlers = [String : [RouterParameterHandler]]()
+
     /// Initialize a `Router` instance
     /// - parameter mergeParameters: Specify if this router should have access to path parameters
     /// matched in its parent router. Defaults to `false`.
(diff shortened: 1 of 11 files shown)


Sources/Kitura/Router.swift
```diff
@@ -157,6 +160,31 @@ public class Router {
         self.all(route, middleware: subrouter)
         return subrouter
     }
+
+    ///
```
shmuelk: Missing API documentation here

Sources/Kitura/Router.swift
```diff
@@ -157,6 +160,31 @@ public class Router {
...
+    @discardableResult
+    public func parameter(_ name: String, handler: @escaping RouterParameterHandler...) -> Router {
+        return self.parameter([name], handlers: handler)
+    }
+
+    ///
```
shmuelk: Missing API documentation here

Sources/Kitura/Router.swift
```diff
@@ -157,6 +160,31 @@ public class Router {
...
+    @discardableResult
+    public func parameter(_ names: [String], handler: @escaping RouterParameterHandler...) -> Router {
+        return self.parameter(names, handlers: handler)
+    }
+
+    ///
```
shmuelk: Missing API documentation here

Sources/Kitura/RouterParameterWalker.swift
```diff
@@ -0,0 +1,57 @@
...
+
+    init(handlers: [String : [RouterParameterHandler]]) {
+        self.parameterHandlers = handlers
+    }
+
+    func handle(request: RouterRequest, response: RouterResponse, with callback: @escaping () -> Void) {
+        let filtered = request.parameters.filter { (key, _) in self.parameterHandlers.keys.contains(key) }
+        self.handle(filtered: filtered, request: request, response: response, with: callback)
+    }
+
```
shmuelk: It took a while to understand what was going on here.

If I understand correctly, you are modifying the copied arrays of parameters and parameterHandlers in order to do the recursion and/or looping without an index (i.e. other state in the RouterParameterWalkser class.

I have two concerns here:
    1. Performance. Which performs better the algorithm in the RouterElementWalker class, which uses an index, albeit captures a local variable or this one which is mdifying, thus copying the arrays.
    2. Consistency. I would have liked all of the "walker" classes to be similar in nature. It makes it easier to maintain them.

shmuelk:
In addition to the changes requested, please update your branch from master.....

naithar:
@shmuelk updated based on review

Sources/Kitura/RouterParameterWalker.swift
```diff
@@ -0,0 +1,67 @@
...
+    /// Collection of 'RouterParameterHandler' for specified parameter name
+    private var parameterHandlers: [String : [RouterParameterHandler]]
+
+    init(handlers: [String : [RouterParameterHandler]]) {
+        self.parameterHandlers = handlers
+    }
+
+    /// Invoke all possible parameter handlers for request
+    ///
+    /// - Parameter request: A current `RouterRequst` that is handled by a server
```
ianpartridge: Typo - `RouterRequest`

Sources/Kitura/RouterParameterWalker.swift
```diff
@@ -0,0 +1,67 @@
...
+    /// - Parameter response: A current `RouterResponse` that is handled by a server
+    /// - Parameter callback: A callback that will be invoked when all handlers are invoked
```
ianpartridge: Should this be "**after** all handlers are invoked"?

Sources/Kitura/Router.swift
```diff
@@ -53,6 +53,9 @@ public class Router {
     /// Flag to enable/disable access to parent router's params
     private let mergeParameters: Bool
 
+    ///
```
ianpartridge: Documentation needed.

Sources/Kitura/Router.swift
```diff
@@ -157,6 +160,46 @@ public class Router {
         self.all(route, middleware: subrouter)
         return subrouter
     }
+
+    /// Handle an URL parameter
```
ianpartridge: Can these comments please explain WHY someone would use this functionality? What is the purpose of the feature / what problem does it solve?

Sources/Kitura/RouterElementWalker.swift
```diff
@@ -36,12 +36,18 @@ class RouterElementWalker {
     /// Copy of request parameters to enable mergeParams
     private let parameters: [String: String]
 
-    init(elements: [RouterElement], request: RouterRequest, response: RouterResponse, callback: @escaping () -> Void) {
+    ///
```
ianpartridge: Comment needed.

Sources/Kitura/RouterParameterWalker.swift
```diff
@@ -0,0 +1,67 @@
...
+            callback()
+            return
+        }
+
+        let filtered = request.parameters.filter { (key, _) in self.parameterHandlers.keys.contains(key) }
+        self.handle(filtered: filtered, request: request, response: response, with: callback)
+    }
+
+    private func handle(filtered: [(String, String)], request: RouterRequest, response: RouterResponse, with callback: @escaping () -> Void) {
+        if filtered.count > 0 {
```
ianpartridge: I would use the same pattern as in the other `handle()` function:

```swift
guard filtered.count > 0 else {
    callback()
    return
}

/// rest of function
```

Tests/KituraTests/TestRequests.swift
```diff
@@ -26,7 +26,8 @@ class TestRequests: XCTestCase {
         return [
                    ("testURLParameters", testURLParameters),
                    ("testCustomMiddlewareURLParameter", testCustomMiddlewareURLParameter),
-                   ("testCustomMiddlewareURLParameterWithQueryParam", testCustomMiddlewareURLParameterWithQueryParam)
+                   ("testCustomMiddlewareURLParameterWithQueryParam", testCustomMiddlewareURLParameterWithQueryParam),
+                   ("testParameters", testParameters)
```
ianpartridge: You can leave a trailing comma here so future tests can be added without modifying this line.

Tests/KituraTests/TestRequests.swift
```diff
@@ -123,4 +124,40 @@ class TestRequests: XCTestCase {
 
         return router
     }
+
+    func testParameters() {
```
ianpartridge: Can we expand the tests a little more? I feel like this is just the basic "hello world" test and there must be more complicated examples we should cover as well.

naithar:
@ianpartridge updated comments and added tests. Not sure if there is more tests to add.

naithar:
There is also another thing i would like to make clear for myself. Currently `RouterParameterWalker` is performed after `RouterElementWalker` performs all it's handlers. [This commit](https://github.com/naithar/Kitura/commit/638811974f031db24bafdce49027d33c91dd54e0) fixes this, but is it actually an issue to be addressed?

shmuelk:
@naithar Do you mean The RouterElementWalker invokes the RouterParameterWalker after all of the routes have been processed (checked for match.....) ?

If so, I think that would be a problem. Every time a RouterElement matches a path, the set of request parameters gets reset to just those in the request. The set may also include parameters from parent routers  if the mergeParameters option was set.

I guess the RouterParameterWalkser needs to be invoked every time a RouterElement has a match and invokes the RouterMiddlewareWalker.

naithar:
@shmuelk currently `RouterParameterWalker` called before parameters are reset, but after element and it's middlewares is processed. In branch that i linked earlier (issue_295-2), i moved `RouterParameterWalker` call right after parameters have been set but before middlewares start handling.

naithar:
I can merge it to this PR for review.

shmuelk:
@naithar That's what I thought I saw in the code. But I wasn't sure after your comment.

The real issue between calling the RouterParameterWalkser before the handlers get called, is what happens when people don't cal the various callbacks. In theory in a RouteHandler, one doesn't have to call net(), as long as one calls end() on the response. The same is true if one doesn't call the callback in a RouterParameterHandler. Maybe that's just life....

naithar:
[This test](https://github.com/naithar/Kitura/blob/638811974f031db24bafdce49027d33c91dd54e0/Tests/KituraTests/TestRequests.swift#L208-L220) is made to check the correct response handling when user doesn't call `next` in `RouterParameterHandler`, but calls `.end()` on response. If i'm not mistaken the result should be the same as replacing it with just:

```swift
        router.get("users/:user/:id") { request, response, next in
            XCTAssertNotNil(request.parameters["id"])
            guard Int(request.parameters["id"]!) != nil else {
                try response.status(.notAcceptable).end()
                return
            }

            response.headers["User-Id"] = request.parameters["id"]
            response.status(.OK).send(data: "\(request.parameters["id"]!)".data(using: .utf8)!)
            next()
        }
```

Which is checked and working.

shmuelk:
@naithar Excellent.

naithar:
@shmuelk merged. Travis Linux build is currently failing, but it's not related to PR as far as i understand

shmuelk:
@naithar We know about the Travis-CI problem. I reported it to them.

Issue #857, opened 2016-11-16
Decode percent encoded string parameters


ianpartridge:
Feedback received on Slack:

>Hello! When using request.queryParameters with GET requests, or using BodyParser/URLEncodedBodyParser for POST, is there a preferred way to get%20rid%20of%20encoding? removingPercentEncoding? Something else?

It would be nice if we provided some functionality to make this automatic.

twostraws:
It's worth adding that using `removingPercentEncoding()` is insufficient to get full coverage. That method chokes on some characters if the form was not submitted with a UTF-8 charset, and also does not replace "+" with " ".

vadimeisenbergibm:
Using the following node.js express code to check how "+" and "%20" are treated in different URL parts:
```javascript
var express = require('express');
var app = express();

app.get('/say hello', function (req, res) {
    res.send("Hello world with whitespace!\n");
});

app.get('/say%20hello', function (req, res) {
    res.send("Hello world with %20!\n");
});

app.get('/say+hello', function (req, res) {
    res.send("Hello world with +!\n");
});

app.get('/hello/:user', function (req, res) {
    res.send("Hello " + req.params.user + "!, the spouse is " + req.query.spouse + "\n");
});

app.listen(3000, function () {
    console.log('Listening on port 3000...');
});
```
```
$ curl "http://localhost:3000/say%20hello"
Hello world with %20!
$ curl "http://localhost:3000/say hello"
curl: (52) Empty reply from server
$ curl "http://localhost:3000/say+hello"
Cannot GET /say+hello
$ curl "http://localhost:3000/hello/John%20Doe?spouse=Jane%20Roe"
Hello John Doe!, the spouse is Jane Roe
$ curl "http://localhost:3000/hello/John%20Doe?spouse=Jane+Roe"
Hello John Doe!, the spouse is Jane Roe
$ curl "http://localhost:3000/hello/John+Doe?spouse=Jane+Roe"
Hello John+Doe!, the spouse is Jane Roe
$ curl "http://localhost:3000/hello/John+Doe?spouse=Jane Roe"
curl: (52) Empty reply from server
```
From the example, and as explained [in this stack overflow answer](http://stackoverflow.com/a/1634293/6028420)

1. For routes, no decoding of "%20" or "+" is performed
2. For path parameters, decoding of "%20" is performed
3. For query parameters, decoding of both "%20" and "+" is performed

vadimeisenbergibm:
Other percentage encodings:
[...]

Issue #830, opened 2016-11-02
Kitura Request.URL should contain Host information


carlbrown:
The Kitura RouterRequest type has a `URL` parameter, but it only contains the request path, so it's not actually a URL (it should also be of `URL` type, but I'm guessing that ship has sailed).

The request's URL should be populated with information from the `Host` header from the HTTP/1.1 request made by the client.

PR #859, opened 2016-11-17
Include full URL in RouterRequest and deprecate url Strings and URLParser in favour of URLComponents

na-gupta:
## Description
Include full URL in RouterRequest and deprecate url Strings and URLParser in favour of URLComponents.

Dependent on IBM-Swift/Kitura-net#142

Addresses IBM-Swift/Kitura#830

This also partially addresses IBM-Swift/Kitura#857 as URLComponents has percent encoded and decoded vars for query params and other segments.

## Checklist:

…URLComponents including scheme and host and corresponding changes


Sources/Kitura/RouterRequest.swift
```diff
@@ -102,13 +93,11 @@ public class RouterRequest {
...
-    public var originalURL: String {
-        return serverRequest.urlString
-    }
+    /// The URL from the request if properly received
+    public internal(set) var url : URL?
 
-    /// The URL.
-    public let url: String
+    /// The URL from the request as URLComponents
+    public internal(set) var urlComponents = URLComponents()
```
ianpartridge: Doesn't this break public API?

na-gupta: Yes, it does. If we wanted to avoid breaking changes, we could use a different name for the URL variable - `xyz: URL`, and deprecate `url: String` and `originalURL: String`, but keep them around for now. I wanted Shmuel and you to provide direction for this. Keep in mind, those variables are broken today as they return just the path of the URL(no scheme or host:port), and not the full URL as they indicate.

na-gupta: A compromise may be that we keep `originalURL: String` around, and just change `url: String` to `url: URL` so that's the only breaking change.

ianpartridge: I think we just add a new field for `public internal(set) var urlComponents = URLComponents()` and keep the old one until 2.0?

na-gupta: That sounds good to me. 

So to summarize - add `urlComponents: URLComponents`, deprecate `url: String` and `parsedURL: URLParser` in favor of urlComponents, and remove internal use of the deprecated variables.

And do we want to use the same approach in Kitura-net as well?

@shmuelk @ianpartridge : Could you confirm this sounds ok?

ianpartridge: I think that sounds good. We need to be confident that `URLComponents` in Foundation is up to scratch.

na-gupta: We do want to fix the deprecated variables (url, urlString, parsedURL) so they contain the full URL, not just the path like they do currently, correct? This can potentially break code that is relying on these variables to only contain the path, but really those variables are currently broken as they do not return what they claim.

Sources/Kitura/RouterRequest.swift
```diff
@@ -130,7 +128,30 @@ public class RouterRequest {
...
+    public lazy var queryParameters: [String:String] = { [unowned self] in
+        var qParams: [String:String] = [:]
+        /*
+        if let queryItems = self.urlComponents.queryItems {
+            for item in queryItems {
+                qParams[item.name] = item.value
+            }
+        }
+        */
+        // urlComponents.queryItems above crashes on linux as of swift 3.0.1
```
ianpartridge: Do we have an open swift.org bug for this?

na-gupta: @ianpartridge  I believe it is related to https://bugs.swift.org/browse/SR-2462
And I think apple/swift-corelibs-foundation#719 contains a workaround for that bug addressing this specific property.

Sources/Kitura/Router.swift
```diff
@@ -239,14 +236,14 @@ extension Router : RouterMiddleware {
 
         let index = urlPath.index(urlPath.startIndex, offsetBy: mountpath.characters.count)
 
-        request.parsedURL.path = urlPath.substring(from: index)
```
shmuelk: The following changes are API breaking changes as you aren't maintaining the old parsedURL.path contents. The idea was to deprecate the old stuff.....

This is true with the whole set of similar changes here....

The fact that parsedURL is now a computed variable doesn't help, as you made it a lazy computed variable. I don't think it gets recomputed on every access.

Sources/Kitura/RouterRequest.swift
```diff
@@ -103,12 +98,21 @@ public class RouterRequest {
...
 
     /// The original URL as a string.
-    public var originalURL: String {
-        return serverRequest.urlString
-    }
+    /// This contains just the path and query parameters starting with '/'
+    /// Use "urlComponents" for the full URL
+    @available(*, deprecated, message:
+        "This contains just the path and query parameters starting with '/'. use 'urlComponents' instead")
+    public var originalURL : String { return serverRequest.urlString }
```
shmuelk: This isn't quite true. urlComponents will change through a request as parts of the path are dropped off in sub-routers. originalURL is the complete original URL before anything happened.

shmuelk:
Also please fix the conflicts as noted by github.

na-gupta:
@shmuelk Requested changes made. We will need to re-run the travis build after IBM-Swift/Kitura-net#142 is merged.

PR #886, opened 2016-11-28
Attempt to fix #879: Once we've called the lifecycle methods, clear t…

### LINKED ISSUE #879: Memory Issues in Kitura Middleware
I'm seeing some issue(s) with Kitura Middleware memory.  On a simple Kitura project with StaticFileServer, BodyParser, Session and Compression, but with no user-supplied Swift code, reloading a static page over and over causes memory to grow without bound.

I've created a sample project at https://github.com/carlbrown/KituraPageLeakSpike which has the [trace from a run using Xcode Instruments](https://github.com/carlbrown/KituraPageLeakSpike/blob/master/LeakInfo/Instruments_Run.trace.zip) and [a video](https://github.com/carlbrown/KituraPageLeakSpike/blob/master/LeakInfo/Instruments_Run.mp4) that shows the problem being exercised. (It uses static content copied from Kitura.io).

Here is the memory footprint from repeatedly hitting "reload":
instrumentsscreenshotcropped

I don't see this issue on the [Kitura Sample](https://github.com/IBM-Swift/Kitura-Sample/) project, so it's somewhere else. Maybe Session?

carlbrown:
I don't know that this is the way we want to go, but this dramatically helps https://github.com/IBM-Swift/Kitura/issues/879 on my machine.

`swift test` passes for me (at least on Darwin).  I'm expecting that creating a Pull Request will kick off a Travis build to test it on Linux, too.

### DIFF
--- a/Sources/Kitura/RouterResponse.swift
+++ b/Sources/Kitura/RouterResponse.swift
@@ -120,6 +120,9 @@ public class RouterResponse {
             try response.write(from: content)
         }
         state.invokedEnd = true
+        
+        //Now that we've called the lifecycle methods, clear them to avoid retain cycles
+        self.lifecycle = Lifecycle()
         try response.end()
     }
 


djones6:
@carlbrown thanks for the PR, we reached the same conclusion (that the references within the Lifecycle are causing a retain cycle), and this is essentially the direction that we were heading (resetting the Lifecycle's state at the end).
@shmuelk may want to comment on how we go about achieving this.

shmuelk:
While this indeed works. I don't like it, because it assumes that all of the lifecycle functions end their "life" at the same time. Granted that this is true now, this might not be true in the future if added more such lifecycle functions.

I'd rather have an implementation where the last one in the chain resets itself to the original "empty function".

carlbrown:
@shmuelk I agree.  It's definitely not clean, completely unintuitive, and I'm concerned about potential side-effects (although I haven't been able to trigger any to-date).

At the moment, though, this hack is what I'm running in "production" until there's a better fix.

shmuelk:
I wrote PR https://github.com/IBM-Swift/Kitura/pull/911 as an alternative to this PR, which has been merged.

PR #887, opened 2016-11-29
Add call site line number to XCTestExpectation description and…

na-gupta:
## Description
- Add call site line number to XCTestExpectation description
- Skip tests if server start fails
- Change osx image for travis to 8.1

## Motivation and Context
When a test makes several calls to performServerTest and one or more expectations timeout, it can be impossible to tell which one failed. By combining the line number of the code that calls performServerTest along with the async task index, we should always be able to narrow the failure to a specific task.

If the server start fails, there is no point in creating and waiting for expectations that will never be fulfilled, so fail right away with the server error.

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ matrix:
       dist: trusty
       sudo: required
     - os: osx
-      osx_image: xcode8
+      osx_image: xcode8.1
       sudo: required
 
 before_install:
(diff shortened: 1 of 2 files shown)


na-gupta:
@shmuelk @ianpartridge Please review. (travis osx builds for this may fail till IBM-Swift/Kitura-net#149 is merged.)

Issue #883, opened 2016-11-27
HTTP Cookie does not handle multiple equals symbols


drewag:
The current implementation requires that a cookie only have a single '=' in it. Instead, it should split the cookie based on the first '=' and everything after that, including any extra equals symbols should be included in the value.

PR #899, opened 2016-12-01
IBM-Swift/Kitura#883 Rewrite cookie parsing implementation

na-gupta:
## Description
- Rewrite Kitura cookie parsing implementation
- Cookie delimiter should be semicolon followed by zero or more whitespace characters
- Use first equals in cookie as name/value delimiter, remaining are treated as part of cookie value
- If cookie value is quoted(starts and ends with `"`), remove quotes
- Ensure unicode characters in cookies values are supported
- Add tests for all these conditions
- Cleanup: remove NSString casts that are no longer needed

## Motivation and Context
Rewrite Kitura cookie parsing implementation to better match standards
Addresses IBM-Swift/Kitura#883 among other things

## Checklist:


Tests/KituraTests/TestCookies.swift
```diff
@@ -61,25 +52,38 @@ class TestCookies: XCTestCase {
 
     let router = TestCookies.setupRouter()
 
-    func testCookieToServer() {
+    func testCookieToServerWithColonSeparator() {
```
ianpartridge: Shouldn't this be `testCookieToServerWithSemiColonSeparator()`?

Sources/Kitura/RouterRequest.swift
```diff
@@ -234,42 +225,46 @@ public class RouterRequest {
...
-        guard let rawCookies = headers[cookieHeader] else {
-            return
+    private static var separator: RegularExpressionType = {
+        do {
+            // matches that do not contain semicolons and do not start with whitespaces
+            // effectively splits string by ";\\s*"
+            return try RegularExpressionType(pattern: "[^;\\s][^;]*", options: [])
+        } catch { // should never throw here, famous last words
+            Log.error("Error creating cookie separator regex: \(error)")
+            exit(1)
```
ianpartridge: Surely this is bad...

Sources/Kitura/RouterRequest.swift
```diff
@@ -234,42 +225,46 @@ public class RouterRequest {
...
-                #endif
-                let theCookie = HTTPCookie(properties:
-                                                [HTTPCookiePropertyKey.domain: ".",
-                                                 HTTPCookiePropertyKey.path: "/",
-                                                 HTTPCookiePropertyKey.name: cookieName ,
-                                                 HTTPCookiePropertyKey.value: cookieValue])
-
-                cookies[cookieNameValueParts[0]] = theCookie
-            }
+    private static func getCookie(cookieString: String) -> HTTPCookie? {
```
ianpartridge: I think including the type name in the variable name is not best practice.

Sources/Kitura/RouterRequest.swift
```diff
@@ -125,15 +121,10 @@ public class RouterRequest {
     public var remoteAddress: String { return serverRequest.remoteAddress }
 
     /// Parsed Cookies, used to do a lazy parsing of the appropriate headers.
-    private lazy var _cookies: Cookies = { [unowned self] in
-        return Cookies(headers: self.serverRequest.headers)
+    public lazy var cookies: [String: HTTPCookie] = { [unowned self] in
```
ianpartridge: Is there a reason why the private `_cookies` has gone?

Sources/Kitura/RouterRequest.swift
na-gupta: Ok, will change it

Tests/KituraTests/TestCookies.swift
ianpartridge: Thanks. It obviously applies to the other tests too.

Sources/Kitura/RouterRequest.swift
na-gupta: I've addressed the other comments, but not sure if there's a better pattern/way to achieve this one?
The pattern literal we're using is pretty straightforward, so any changes to RegularExpression to break this would have to be really bad.

ianpartridge: Crashing Kitura cannot be right...  Also, why do you create a new `RegularExpressionType` every time this computed property is accessed? Compiling the regex is likely to have a non-trivial cost so should only be done once.

ianpartridge:
I've reviewed this but I would like @shmuelk to take a good look at the changes to cookie handling in case there's anything I've missed.

Sources/Kitura/RouterRequest.swift
ianpartridge: OK, I don't mind removing it, @shmuelk may have had a reason the code was structured this way though.

Sources/Kitura/RouterRequest.swift
na-gupta: Not sure what you mean, I'm only creating the regex once. It is not computed each time.
`private static var separator: RegularExpressionType = {...}()`

na-gupta: I can make the separator regex optional, and let the cooking parsing return an empty dictionary in the unlikely case where it fails. Does that sound like a better approach?

ianpartridge: OK, maybe I'm going mad here, but just because it's `static` doesn't mean it's only computed once. Every time the property is accessed it is recomputed.

Try running this:

```swift
class Test {
    static var aString: String {
        print("test")
        return "string"
    }
}

print(Test.aString)
print(Test.aString)
```

It prints:

```
test
string
test
string
```

ianpartridge: Yes, the optional idea sounds better to me. We need a solution that doesn't bring down the server.

na-gupta: I'm not using a computed value. The code I use is initializing it from a closure return just as if I did:
`private static var separator: RegularExpressionType = createRegex()`

It's similar to:
```
class Test {
    static var aString: String = {
        print("test")
        return "string"
    }()
}

print(Test.aString)
print(Test.aString)
```
For which the output is:
```
test
string
string
```

ianpartridge: !! I totally missed the closure execution. Apologies.

na-gupta:
Remove regex use per discussion with @shmuelk 
Also added support to unquote cookie values as per spec and added tests for it
@shmuelk @ianpartridge Could you please review this again.

Sources/Kitura/RouterRequest.swift
```diff
@@ -234,42 +225,41 @@ public class RouterRequest {
...
+        var cookies = [String: HTTPCookie]()
+        guard let cookieHeaders = headers["cookie"] else {
+            return cookies
         }
-        for cookie in rawCookies {
-            initCookie(cookie, cookies: &cookies)
+
+        for cookieHeader in cookieHeaders {
+            for cookie in cookieHeader.components(separatedBy: ";") {
+                let trimmedCookie = cookie.trimmingCharacters(in: .whitespaces)
```
ianpartridge: We didn't trim this before, did we? Is this specified in an RFC or something?

na-gupta: The standard for the separator is semicolons with zero or more whitespace. Shmuel suggested removing the regex, splitting by semicolons and stripping out leading/trailing whitespace, which achieves the goal without using regular expressions.

ianpartridge:
This looks good to me now. @shmuelk merge if you're happy?

shmuelk:
Sorry.

shmuelk:
Looks ok. I'll merge after CI completes.

PR #911, opened 2016-12-06
Eliminate potential reference loops within RouterResponse lifecycle functions

shmuelk:
## Description
Kitura provides in RouterResponse life cycle hooks, i.e. places where functions/closures can be invoked during the life cycle of the request. In some cases the closures used to implement these hooks cause object reference loops. This in turn causes memory leaks. 

This PR provides a set of functions to reset the life cycle functions each back to their initial value, a no-op function. This breaks the reference loop and fixes the memory leak.

## Motivation and Context
Fix issue IBM-Swift/Kitura#879

## How Has This Been Tested?
@djones6 Ran memory leak tests and verified the leaks went away.
I also ran the Kitura unit tests on both macOS and Linux.

## Checklist:

### DIFF
--- a/Sources/Kitura/RouterResponse.swift
+++ b/Sources/Kitura/RouterResponse.swift
@@ -34,10 +34,10 @@ public class RouterResponse {
         var invokedSend = false
     }
 
-    /// A set of functions called durin the life cycle of a Request.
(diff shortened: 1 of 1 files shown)


na-gupta:
👍

PR #913, opened 2016-12-06
Added test for isContentType()

kweinmeister:
## Description
Added test for isContentType()

## Motivation and Context
No test for this function yet

## How Has This Been Tested?
Tests passed

### DIFF
--- a/Tests/KituraTests/TestContentType.swift
+++ b/Tests/KituraTests/TestContentType.swift
@@ -23,34 +23,32 @@ class TestContentType: XCTestCase {
 
     static var allTests: [(String, (TestContentType) -> () throws -> Void)] {
         return [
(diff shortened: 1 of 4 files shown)


Tests/KituraTests/TestContentType.swift
```diff
@@ -71,4 +69,22 @@ class TestContentType: XCTestCase {
         XCTAssertEqual(result, "text/html")
     }
 
+    func testIsContentType() {
+        var result = contentType.isContentType("application/json", ofType: "json")
```
shmuelk: You didn't add this test to the allTests variable up on top.

When unit testing tests please double check that the number of tests run on Linux equals the number of tests run on macOS.

PR #926, opened 2016-12-09
Break request/response retain cycle. Intended to help #879

carlbrown:
Break strong retain cycles in the process() method

## Description
Code should be pretty self explanatory.
Graph changes below generated with this fix and https://github.com/IBM-Swift/Kitura-Compression/pull/26 and https://github.com/IBM-Swift/Kitura-Session/pull/29

## Motivation and Context
See leak graph posted in #879 

## How Has This Been Tested?
Test suite still passes.  In running on instruments, I get fewer leaks in "Cycles and Roots".

Before: 
screen shot 2016-12-09 at 10 51 21 am

After:
screen shot 2016-12-09 at 10 51 06 am

### DIFF
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -268,11 +268,23 @@ extension Router : ServerDelegate {
         let routeReq = RouterRequest(request: request)
         let routeResp = RouterResponse(response: response, router: self, request: routeReq)
 
-        process(request: routeReq, response: routeResp) { [unowned self] () in
+        process(request: routeReq, response: routeResp) { [weak self, weak routeReq, weak routeResp] () in
+            guard let strongSelf = self else {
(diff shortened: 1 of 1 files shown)


Sources/Kitura/Router.swift
```diff
@@ -268,11 +268,23 @@ extension Router : ServerDelegate {
         let routeReq = RouterRequest(request: request)
         let routeResp = RouterResponse(response: response, router: self, request: routeReq)
 
-        process(request: routeReq, response: routeResp) { [unowned self] () in
+        process(request: routeReq, response: routeResp) { [weak self, weak routeReq, weak routeResp] () in
+            guard let strongSelf = self else {
+                Log.error("Found nil self at #file #line")
```
carlbrown: Yeah, you're right.

Issue #908, opened 2016-12-05
BodyParser does not pass body data to request's body if not text


jacobvanorder:
As raised in [this PR](https://github.com/rfdickerson/Kitura-Protobufs/pull/2) for @rfdickerson's protobuf example, `BodyParser` is not retaining the body data made by a request if it isn't a [type outlined](https://github.com/IBM-Swift/Kitura/blob/master/Sources/Kitura/bodyParser/BodyParser.swift#L31-L34) in `BodyParser`'s `parserMap` dictionary. 

In this case, the `request.body` was always `.none` instead of the expected type `.raw(Data)` [from `ParsedBody`](https://github.com/IBM-Swift/Kitura/blob/1249f60a27e7f19a55b5569cd3b630312f67b881/Sources/Kitura/bodyParser/ParsedBody.swift#L42).

As it stands, I'm not sure if anything except what is outlined in `parserMap` gets retained but `application/octet-stream` sure wasn't.

naithar:
If i'm not mistaken `getParser` method in [BodyParser.swift](https://github.com/IBM-Swift/Kitura/blob/master/Sources/Kitura/bodyParser/BodyParser.swift) determines which parser to use.  
Atm it handles types included in `parserMap`, all `text` content-types and `multipart/form-data`.  
So other types, including `application/octet-stream` is not parsed.

jacobvanorder:
What is the point of [.raw(Data)](https://github.com/IBM-Swift/Kitura/blob/1249f60a27e7f19a55b5569cd3b630312f67b881/Sources/Kitura/bodyParser/ParsedBody.swift#L42) then? 

In @rfdickerson's example he used `BodyParser` for `JSON` payloads (rightly so). He assumed, as I would have, that because there was `.raw(Data)`, it would default to that which he could use for `application/octet-stream`. It didn't. Instead, it `nil`'d out the data which, in my opinion, seems destructive.

shmuelk:
@jacobvanorder You are probably correct in that we should at least enable the use of .raw(Data) if the mime type isn't one of those that we know how to parse.

It's an interesting question as what is the correct thing to do. One might say, if you don't know what to do, then do nothing. Let the developer read it in themselves. Or use .raw(Data).

jacobvanorder:
[...]

PR #927, opened 2016-12-12
Added RawBodyParser for BodyParser default parsing method (#908).

jacobvanorder:
`BodyParser` should default a `ParsedBody.raw(Data)` when the `Content-Type` isn't known.

## Description
`BodyParser` has a [`parserMap`](https://github.com/IBM-Swift/Kitura/blob/master/Sources/Kitura/bodyParser/BodyParser.swift#L31-L34) which designates that text-based `Content-Type`s correspond to a specialized parser (e.g., `application/json` -> `JSONBodyParser`).

If the `Content-Type` was not any of those types, it would fall-through and `nil` out the data from the `RouterRequest` instead of storing the raw data in [`.raw(Data)`](https://github.com/IBM-Swift/Kitura/blob/master/Sources/Kitura/bodyParser/ParsedBody.swift#L42) on `ParsedBody`.

I fixed this by following pattern and creating `RawBodyParser` which just slots the data into `.raw(Data)`. `BodyParser` now defaults to passing the RouterRequest body into `RawBodyParser`.

## Motivation and Context
This is in response to [Issue 908](https://github.com/IBM-Swift/Kitura/issues/908). While exploring @rfdickerson's [Kitura Protobuf Sample](https://github.com/IBM-Swift/Kitura-Protobuf-Sample), I notice that `POST`ing did not work when sending a protobuf file. The `RouterRequest`'s body was `nil` not `application/json`. @rfdickerson [assumed](https://github.com/IBM-Swift/Kitura-Protobuf-Sample/commit/0475ab18045593e9d5a0b9fff511ce4c5adf2f8d#diff-c59d9b61114ad3468cfbb3a6a44fa297L59), as I would have, that the body could have been accessed using `.raw(Data)` on the `ParsedBody`. This was not the case.

## How Has This Been Tested?

I wrote a unit test that tests whether raw data being sent passes through the `router` properly and can be accessed using `.raw(Data)`.

I only ran tests locally on Mac OS X since I do not have access to a Linux environment immediately.

This would only have impact on existing code if the end user _knew_ that `BodyParser` would `nil` out the body and wrote against that scenario.

## Checklist:

### DIFF
--- a/Sources/Kitura/bodyParser/BodyParser.swift
+++ b/Sources/Kitura/bodyParser/BodyParser.swift
@@ -95,8 +95,9 @@ public class BodyParser: RouterMiddleware {
                 boundary = boundary.substring(to: parameterStart.lowerBound)
             }
             return MultiPartBodyParser(boundary: boundary)
+        } else { //Default: parse body as `.raw(Data)`
+            return RawBodyParser()
         }
-        return nil
     }
 
     /// Read incoming message for Parse
--- a/Sources/Kitura/bodyParser/RawBodyParser.swift
+++ b/Sources/Kitura/bodyParser/RawBodyParser.swift
@@ -1,0 +1,23 @@
+/*
+ * Copyright IBM Corporation 2016
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import Foundation
+
+class RawBodyParser: BodyParserProtocol {
+    func parse(_ data: Data) -> ParsedBody? {
+        return ParsedBody.raw(data)
+    }
+}
--- a/Tests/KituraTests/TestResponse.swift
+++ b/Tests/KituraTests/TestResponse.swift
@@ -320,7 +320,40 @@ class TestResponse: XCTestCase {
                 req.write(from: "This does not contain any valid boundary")
             }
         }
-
+        
+    }
+    
+    func testRawDataPost() {
+        performServerTest(router) { expectation in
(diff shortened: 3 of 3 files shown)


jacobvanorder:
Realized the `arc4Random` isn't available on Linux. I'll update shortly.

ianpartridge:
Thanks for your PR! Please don't use random data in the test. It makes recreating any future failures theoretically impossible.

jacobvanorder:
@ianpartridge would a `Data` object filled with zeros be better? The point is that raw data is coming through. Don't care what but just that the length is the same. Should I be testing for something else?

ianpartridge:
If you can test the same thing using just zeroes then please do that :)

jacobvanorder:
@ianpartridge Will do! Thanks!

youming-lin:
All zeros sounds to me like an edge case.... maybe have another test with statically generated random data?

jacobvanorder:
@ianpartridge Updated!

ianpartridge:
Thanks for your work on this, @jacobvanorder!

jacobvanorder:
@ianpartridge I'd love to chip in more. Is there a standard for picking off issues or is it anything that isn't assigned is fair game?

ianpartridge:
Hi @jacobvanorder.  Any issue that isn't assigned is available - let us know if a particular one looks interesting.  Is there a particular area of Kitura that you've been investigating or using?  If you are looking for a standalone feature to work on, we'd love `Kitura-Credentials` to support more providers (currently we have Facebook, Google and GitHub).  How about creating `Kitura-Credentials-Twitter` so Kitura apps can provide login through the [Twitter OAuth API](https://dev.twitter.com/oauth/reference/post/oauth2/token)?

jacobvanorder:
@ianpartridge Thanks for the direction! 

I came to this issue after investigating Protocol Buffers and @rfdickerson's example for a talk I'll be doing in the spring. I've been trying to frame the depth of what I get into concerning Kitura balanced with my time limit for the talk.

I'll take a look at Twitter OAuth and contribute how I can. This has been really fun to delve into and I'm happy to work on something new.

ianpartridge:
Sounds good - let us know how you get on!

JohnKnust:
The information to set a custom Content-Type could be helpful. 
I didn't set it and it the parser will return here ;) 
I use now: "custom/row" and it works well :)	

if let parser = parserMap[contentTypeWithoutParameters] {
            return parser

PR #931, opened 2016-12-14
Added test case for uncovered FileResourceServer code, and made minor fixes to FileResourceServer

kweinmeister:
- Added test for missing Kitura resource
- Updated FileResourceServer logic to return `.notFound` HTTP status code when file not found
- Fixed typo: "potental"

### DIFF
--- a/Sources/Kitura/FileResourceServer.swift
+++ b/Sources/Kitura/FileResourceServer.swift
@@ -22,6 +22,11 @@ class FileResourceServer {
     /// if file found - send it in response
     func sendIfFound(resource: String, usingResponse response: RouterResponse) {
         guard let resourceFileName = getFilePath(for: resource) else {
+            do {
+                try response.status(.notFound).end()
+            } catch {
+                Log.error("failed to send not found response")
+            }
             return
         }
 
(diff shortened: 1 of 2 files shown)


shmuelk:
FileServer should NOT return .notFound when a file is not found. You are preventing other RouterHandlers and Middleware from handling the request.

Sources/Kitura/FileResourceServer.swift
```diff
@@ -22,6 +22,11 @@ class FileResourceServer {
     /// if file found - send it in response
     func sendIfFound(resource: String, usingResponse response: RouterResponse) {
         guard let resourceFileName = getFilePath(for: resource) else {
+            do {
+                try response.status(.notFound).end()
+            } catch {
+                Log.error("failed to send not found response")
+            }
```
shmuelk: ~~FileServer should NOT return .notFound when a file is not found. You are preventing other RouterHandlers and Middleware from handling the request.~~

The comments said the change was to FileServer. The was in fact to FileResoureServer. That is a very different story.

na-gupta: As per the [w3 spec](https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html#sec10.4) - "Except when responding to a HEAD request, the server SHOULD include an entity containing an explanation of the error situation, and whether it is a temporary or permanent condition."

We do a similar thing (sending an error message for the body) [here](https://github.com/IBM-Swift/Kitura/blob/master/Sources/Kitura/Router.swift#L342-L343)

kweinmeister:
OK, I will add a response body and ensure that the test checks for a body.

kweinmeister:
@na-gupta Updated implementation and test, per your comments.

@shmuelk Updated title of PR.

Issue #942, opened 2016-12-21
Performance regression with urlURL


djones6:
The changes for the `RouterRequest` URL API (#859 and #904, and the associated changes in Kitura-net) caused a throughput regression.

The original change to use `URLComponents` (#859) costs around 6% throughput, but also introduced a memory leak (a bug in Foundation), which #904 resolves using `URL`.  Unfortunately, it turns out that creating a `URL` is expensive - this cost us over 20%, adding up to around 30% regression overall on our 1.2.0 performance (for simple hello-world type requests).

Ultimately the solution is URLComponents, once the memory leak is fixed.  Until then, I'd like to avoid the ~30% performance regression when possible.

shmuelk:
@djones6 Please see KituraNet branch issue_933_2 which attempts to defer the creation of the URLs as much as possible.

djones6:
On an Ubuntu 16.04 box with Swift 3.0.2,  Kitura 1.2.0 (pre-URL changes) `/plaintext` performance = 67k req/s.
1.4.0 (plus check for log level before accessing headers in `onHeadersComplete()`) = 48k
Calculation of the full url (as a `String`) but not creating a `URL` = 63k
Calculation of the full url and creating a `URL` (but not using it in `RouterRequest`) = 54k
Using the `URL` in `RouterRequest` (getting the `String` representation and converting it to a `Data`) = 48k

In other words, the costs:
- Accessing the headers and building up the URL string = 6%
- Creating the URL object = 14%
- Converting the URL object to a Data so that we can pass it to URLParser = 9%

The lazy initialization introduced in IBM-Swift/Kitura-net#162 and IBM-Swift/Kitura#943 recovers the whole regression for cases where the `URL` does not have to be initialized.  Reasons why we may still initialize it include accessing the query params, the `parsedURL` property of `RouterResponse`, or the `urlURL` property of `ServerRequest`.

The ~6% regression that I measured in Kitura 1.2.2, when `URLComponents` was introduced, may well have been the cost of calculating the full url - ie. not the cost of creating a `URLComponents`.  It should still be our long-term goal to move back once the memory leak is resolved.

djones6:
For reference, I raised https://bugs.swift.org/browse/SR-3481 for the memory leak in URLComponents.

PR #943, opened 2016-12-22
Issue 942: Lazy parsedURL for better performance

djones6:
## Description
Lazily construct the `parsedURL` in `RouterRequest`, to avoid the performance penalty of constructing a `URL` in Kitura-Net until we need it (such as retrieving the full URL of a request, or the query params).

For simple cases such as loading static content, we can skip this cost altogether.

A more comprehensive solution would be to avoid the cost of construction for accessing the query params, however since the long-term solution is to use `URLComponents`, this change is sufficient to solve the regression for the most frequent cases without requiring more changes to the public API.

## Motivation and Context
Part of the solution to #942 (the other part is a change to IBM-Swift/Kitura-net, which makes the construction of the `URL` in `HTTPIncomingMessage` lazy).

## How Has This Been Tested?
I ran the Kitura tests on Linux and they passed.  I ran local performance tests to establish that performance recovers to the level we had at Kitura 1.2.0 (as long as query parameters are not accessed).

### DIFF
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -231,8 +231,8 @@ extension Router : RouterMiddleware {
     /// - Parameter next: The closure to invoke to cause the router to inspect the
     ///                  path in the list of paths.
     public func handle(request: RouterRequest, response: RouterResponse, next: @escaping () -> Void) throws {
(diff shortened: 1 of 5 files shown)


shmuelk:
This PR looks very good. It can merged once CI completes

Issue #951, opened 2016-12-29
Memory Leak when checking headers


carlbrown:
Sample project at https://github.com/carlbrown/KituraHeaderLeakSpike 

![kituraheaderleakspike_mem](https://cloud.githubusercontent.com/assets/67206/21553416/c6488e5c-cdcd-11e6-90e0-5b9107423474.png)

I've been chasing what I thought was a leak in Kitura-Compression for days now, but it turns out that's not the issue.  The problem (or at least *_this_* problem, I'm sure there are more) happens when you verify that the Accept-Encoding header exists.

Check https://github.com/carlbrown/KituraHeaderLeakSpike/blob/master/bin/cycle_test.sh for the test that created the above graph (note that the hostname is hardcoded to the test server I'm using).

Have only looked at/for this under Linux (which is what I care about). It may or may not exist under Darwin.

carlbrown:
So the difference between the sections of the graph that are relatively flat and the sections that are near a 45 degree angle is whether or not
```
guard let encodingMethod = req.accepts(header: "Accept-Encoding", types: ["gzip", "deflate", "identity"])
```
gets run.

No idea (yet) why that's causing a leak, but for the record, it was **REALLY** hard to find, because it didn't occur to me to look anywhere near there.

carlbrown:
Near as I can tell, this only happens on Linux.  Here's the same graph with Kitura running on my laptop and cycle_test running from the server (the opposite of the previous graph):

![laptop_mem_graph](https://cloud.githubusercontent.com/assets/67206/21554465/3000be3a-cdd5-11e6-9304-47944a17b0cd.png)

carlbrown:
And that makes this suck, because we won't have access to Instruments to find this, and valgrind is of limited help, because there are *_so, so, very many_* different leaks reported that it's a lot of work to isolate just one to figure out what's going on and test potential solutions.

carlbrown:
OK, so here are the leaks from `valgrind` that are reported when `/startLeaks` has been `POST`ed but aren't reported when it hasn't:
```
_TZFCC6Kitura13RouterRequest16MimeTypeAcceptor7acceptsfT12headerValuesGSaSS_5typesGSaSS__GSqSS_ (MimeTypeAcceptor.swift:67)
[...]

PR #954, opened 2017-01-05
IBM-Swift/Kitura#951 Remove regex for matching accept headers, fix wrong specs ...

na-gupta:
## Description
- Remove `String.range()` call with `.regularExpression` option for matching wildcard accept headers and replace with `hasPrefix()`
- fix multiple incorrect spec implementations
    - Wildcard to accept all types is `*/*` for `Accept` header and `*` for others like `Accept-Charset`  and `Accept-Encoding`
    - quality value of 0 indicates not acceptable and should not be used as a match
    - `application/xml` should match `application/*` but not a pattern like `application.xml`
- add tests for these conditions

## Motivation and Context
See #951

## How Has This Been Tested?
- Test suite on osx and linux
- Additional tests added to cover missing conditions
- Confirm linux memory leak is gone when checking accept headers with type/*

## Checklist:

### DIFF
--- a/Sources/Kitura/MimeTypeAcceptor.swift
+++ b/Sources/Kitura/MimeTypeAcceptor.swift
@@ -62,9 +62,10 @@ extension RouterRequest {
         ///
         /// - Parameter headerValues: Array of Accept header values.
         /// - Parameter types: Array of content/mime type strings.
+        /// - Parameter wildcard: Special value that matches any value. For example "*" or "*/*"
         /// - Returns: Most acceptable type or nil if there are none
-        static func accepts(headerValues: [String], types: [String]) -> String? {
-            let criteriaMatches = getCriteriaMatches(headerValues: headerValues, types: types)
+        static func accepts(headerValues: [String], types: [String], wildcard: String) -> String? {
+            let criteriaMatches = getCriteriaMatches(headerValues: headerValues, types: types, wildcard: wildcard)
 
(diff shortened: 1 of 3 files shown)


Sources/Kitura/RouterRequest.swift
```diff
@@ -208,7 +208,8 @@ public class RouterRequest {
         }
 
         let headerValues = acceptHeaderValue.characters.split(separator: ",").map(String.init)
-        return MimeTypeAcceptor.accepts(headerValues: headerValues, types: types)
+        let wildcard = (header == "Accept" ? "*/*" : "*")
```
shmuelk: You need to do a case insensitive compare.  Header names are case insensitive.

Sources/Kitura/MimeTypeAcceptor.swift
```diff
@@ -62,9 +62,10 @@ extension RouterRequest {
         ///
         /// - Parameter headerValues: Array of Accept header values.
         /// - Parameter types: Array of content/mime type strings.
+        /// - Parameter wildcard: Special value that matches any value. For example "*" or "*/*"
```
shmuelk: I understand what you're trying to do and it's a good fix.

I think the field/parameter needs to renamed some what to reflect that it is a "full value wildcard". Technically a mime type string of text/* is a wildcard of sorts.

na-gupta:
@shmuelk Addressed your comments.

PR #956, opened 2017-01-05
Remove use of String.CompareOptions.regularExpression because of memory leak

na-gupta:
## Description
Replace String.CompareOptions.regularExpression with a static regex.
This has the same memory leak issue described in #951

## Motivation and Context
See #951

## How Has This Been Tested?
With master branch, memory leak reported by massif for 10,000 calls is **155 MB**
This change drops it down to **96 KB** 

## Checklist:

### DIFF
--- a/Sources/Kitura/RouterResponse.swift
+++ b/Sources/Kitura/RouterResponse.swift
@@ -76,6 +76,9 @@ public class RouterResponse {
 
     private var lifecycle = Lifecycle()
 
+    // regex used to sanitize javascript identifiers
+    private static var sanitizeJSIdentifierRegex: RegularExpressionType? = nil
+
     /// Set of cookies to return with the response.
     public var cookies = [String: HTTPCookie]()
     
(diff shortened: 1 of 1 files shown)


Sources/Kitura/RouterResponse.swift
```diff
@@ -238,9 +241,12 @@ public class RouterResponse {
     /// is the alphanumeric characters and `[]$._`).
     /// - Returns: this RouterResponse.
     public func send(jsonp: JSON, callbackParameter: String = "callback") throws -> RouterResponse {
+        if RouterResponse.sanitizeJSIdentifierRegex == nil {
```
shmuelk: This is a bad way to initialize a Static field. It is NOT thread safe.

A better way to do this is is something similar to the way the static field oneTime in KituraNet.ClientRequest is initialized.

na-gupta: Fixed.

na-gupta: @shmuelk I may be wrong, but I think `static private let oneTime = OneTimeInitializations()` in KituraNet.ClientRequest is actually currently broken. 

Swift lazily initializes class variables and constants and since `oneTime` is never referenced anywhere, it never actually gets initialized and so `curl_global_init(Int(CURL_GLOBAL_SSL))` in `OneTimeInitializations.init()` never gets invoked.

Issue #936, opened 2016-12-18
staticFileServer's getAbsolutePath doesn't handle "/" as a path


fim:
I was trying to initialize a StaticFileServer where the path was "/" and my requests kept failing.

I managed to trace the problem down to getAbsolutePath() and the way it handles trailing separators; it simply removes them. This will obviously cause issues when the path is simply "/" since the end result is an empty string. A hacky way to work around this is to specify "//" as the path which will return "/'

A potential fix which I tested and works is:

```diff
diff --git a/Sources/Kitura/staticFileServer/ResourcePathHandler.swift b/Sources/Kitura/staticFileServer/ResourcePathHandler.swift
index bbc1823..d4904d3 100644
--- a/Sources/Kitura/staticFileServer/ResourcePathHandler.swift
+++ b/Sources/Kitura/staticFileServer/ResourcePathHandler.swift
@@ -26,7 +26,7 @@ extension StaticFileServer {

         static func getAbsolutePath(for path: String) -> String {
             var path = path
-            if path.hasSuffix(separator) {
+            if path.hasSuffix(separator) && path != "/" {
                 path = String(path.characters.dropLast())
             }

```

However I think there is a need for some file path normalization routines and I haven't looked to see if this code is used elsewhere so I'm not sure how you want to tackle this. 

Let me know if you have any questions.

Thanks!

youming-lin:
This issue got me thinking about how we are normalizing file paths. Currently, we normalize based on the current directory path, specifically, `FileManager().currentDirectoryPath`. I think this is akin to normalizing paths based on PWD, which means that, the static file is fundamentally different if you change your PWD when running the server app (e.g., `.build/debug/app` vs. `cd .build/debug && ./app`).

Does this actually make sense? For NodeJS, file paths are normalized based on the source file file path; the Swift equivalent would be the `#file` compiler directive. However, since Swift is a compiled language, I think it makes more sense to normalize file paths based on the absolute path of the executable, e.g, `CommandLine.arguments[0]` normalized against PWD.

PR #969, opened 2017-01-17
Allow developer to point static file server to system root. (#936)

joyychang:
## Description

- Added a check for the case when the path input is equal to the separator "/".
- Added a unit test to see the getAbsolutePath function would return "/".

## Motivation and Context
#936 

## How Has This Been Tested?
Change includes a unit test to see that when given the input "/", the function returns "/"

## Checklist:

### DIFF
--- a/Tests/KituraTests/TestStaticFileServer.swift
+++ b/Tests/KituraTests/TestStaticFileServer.swift
@@ -221,4 +221,8 @@ class TestStaticFileServer: XCTestCase {
         runGetResponseTest(path: "/@@Kitura-router@@/missing.file", expectedStatusCode: HTTPStatusCode.notFound)
     }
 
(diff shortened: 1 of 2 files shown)


Tests/KituraTests/TestStaticFileServer.swift
```diff
@@ -221,4 +221,8 @@ class TestStaticFileServer: XCTestCase {
         runGetResponseTest(path: "/@@Kitura-router@@/missing.file", expectedStatusCode: HTTPStatusCode.notFound)
     }
 
+    func testAbsolutePathFunction() {
```
na-gupta: @joyychang This is late as this is already merged, and it's not really a big deal for this one, but in general we should add any new tests to `static var allTests` so they are tested on linux builds as well.

Issue #962, opened 2017-01-09
Add end to end unit tests for SSLConfig


na-gupta:
Currently we are doing end to end tests only for http requests. We need to add comprehensive unit tests for https as well.

Issue #959, opened 2017-01-09
Unable to create SSL context


keshavakarthik:
With  Kitura (.Package(url: "https://github.com/IBM-Swift/Kitura", majorVersion: 1, minor: 4)) adding SSL Support ( Self-Signed ) giving the error
"Error accepting client connection: Error code: 12, ERROR: Unable to create SSL context."

Whereas the same code is working fine with Kitura Version (Package(url: "https://github.com/IBM-Swift/Kitura.git", majorVersion: 1, minor: 0))

na-gupta:
@gtaban @shmuelk 
It looks like this problem was originally introduced with IBM-Swift/Kitura-net#155 in [this commit](https://github.com/IBM-Swift/Kitura-net/commit/93ddb4126ae6555b1856c8dda96e40d197216e16) between Kitura-net 1.3.0 and 1.3.1

On that commit, the problem goes away with commenting out [this line](https://github.com/IBM-Swift/Kitura-net/blob/master/Sources/KituraNet/IncomingSocketHandler.swift#L313) that was added in that commit. However, commenting out that same line on master does not fix the issue. So it looks like some other changes since then makes the problem reappear.

Further investigation is needed to find out how this should be properly fixed.

shmuelk:
@navneet thanks for the research work.

@billabt I did an additional test here, after you pointed out to me that SSL on Kitura 1.2 works and fails on later versions. I commented out line 1305 in Socket.swift (self.delegate?.deinitialize() ). This is in the function deinit of Socket.swift. After commenting out that line of code SSL **works**.

I believe the problem is in the deinitialization of SLLService instances somehow, or somehow they are being shared across Socket instances.

billabt:
Fixed in BlueSocket v0.12.18 and BlueSSLService v0.12.12.

na-gupta:
@billabt Thanks. That fixes the problem for the most part. But in some [new unit tests](https://github.com/IBM-Swift/Kitura/blob/issue_962/Tests/KituraTests/TestStaticFileServer.swift#L58-L60) I added in [branch 962](https://github.com/IBM-Swift/Kitura/tree/issue_962), there is a new failure:
```
xctest(6295,0x70000666b000) malloc: *** error for object 0x7fb5c9653af0: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
```
[...]

PR #973, opened 2017-01-18
State guards for RouterResponse end() and send() functions and use SSL tests

na-gupta:
## Description
- Add state guards for RouterResponse end() and send() functions
- Make most unit tests use SSLService as it surfaces socket read, write, close contention problems whereas Socket mostly ignores them
- We still need some fixes in SSLService before we can enable SSL tests. This just requires us to change the default for `sslOption` in KituraTest.performServerTest() from `SSLOption.httpOnly` to `SSLOption.both`

## Motivation and Context
#959
#962

## Checklist:


shmuelk:
This is ready for merging once travis-ci runs successfully

Issue #940, opened 2016-12-20
Improve performance by using simple string compare of Route paths where possible


shmuelk:

djones6:
Plaintext benchmark results
### Ubuntu 14.04:
```
               | Throughput (req/s)      | CPU (%) | Mem (kb)     | Latency (ms)                   | good
Implementation | Average    | Max        | Average | Avg peak RSS | Average  | 99%      | Max      | iters
---------------|------------|------------|---------|--------------|----------|----------|----------|-------
         1.5.0 |    51787.7 |    52398.1 |    96.1 |        23065 |      2.4 |      2.7 |    198.9 |     3
    +issue_940 |    61069.7 |    62076.0 |    98.3 |        21810 |      2.0 |      2.3 |     19.2 |     3
```
### Ubuntu 16.04:
```
               | Throughput (req/s)      | CPU (%) | Mem (kb)     | Latency (ms)                   | good
Implementation | Average    | Max        | Average | Avg peak RSS | Average  | 99%      | Max      | iters
---------------|------------|------------|---------|--------------|----------|----------|----------|-------
         1.5.0 |    65561.8 |    65926.2 |    95.0 |        21705 |      1.9 |      2.3 |     17.3 |     3
    +issue_940 |    83514.2 |    84434.4 |    98.8 |        23144 |      1.5 |      4.6 |    198.9 |     3

```

djones6:
Results of plaintext with 100 simple non-matching routes ([100Routes](https://github.com/djones6/Swift-TechEmpower/blob/master/kitura/Sources/100Routes/main.swift)) shows even more benefit:
### Ubuntu 16.04:
```
               | Throughput (req/s)      | CPU (%) | Mem (kb)     | Latency (ms)                   | good
Implementation | Average    | Max        | Average | Avg peak RSS | Average  | 99%      | Max      | iters
---------------|------------|------------|---------|--------------|----------|----------|----------|-------
         1.5.0 |     9326.6 |     9952.8 |    89.0 |        23272 |     13.7 |     14.9 |     42.2 |     2
    +issue_940 |    36145.6 |    36270.8 |    87.0 |        21230 |      3.5 |      3.6 |    206.4 |     2
```
### Macbook Pro:
```
               | Throughput (req/s)      | CPU (%) | Mem (kb)     | Latency (ms)                   | good
Implementation | Average    | Max        | Average | Avg peak RSS | Average  | 99%      | Max      | iters
[...]

PR #975, opened 2017-01-19
If the pattern is a simple String, do a String compare and not ReguarExpresion match

shmuelk:
## Description
A check is made on each route pattern to see if it contains a capture group, (:key or (\w) ) or any regular expression matching (*+[]\.) characters. If it doesn't, at run time a String compare for non-partial match Route elements or a String.hasPrefix is done, in lieu of a RegularExpression
match.

## Motivation and Context
Improve performance. On theTechEmpower hello world benchmark perfrmance improves on Ubuntu 14.04 by about 18% and Ubuntu 16.04 by about 27%.

See issue IBM-Swift/Kitura#940

## How Has This Been Tested?
@djones6 Tested to get the performance numbers.
I ran the Kitura unit tests on Linux and macOS

## Checklist:


Sources/Kitura/RouteRegex.swift
```diff
@@ -52,10 +55,17 @@ public class RouteRegex {
...
+    ///            this is a simple String compare, and array of keys
+    internal func buildRegex(fromPattern: String?, allowPartialMatch: Bool = false) -> (RegularExpressionType?, Bool, [String]?) {
         guard let pattern = fromPattern else {
-            return (nil, nil)
+            return (nil, false, nil)
+        }
+        
+        // Check and see if the pattern is a simple string (no captures and not a regular expression)
+        let range = NSMakeRange(0, pattern.characters.count)
+        guard let _ = simpleStringChecker.firstMatch(in: pattern, options: [], range: range) else {
```
vadimeisenbergibm: The following can be achieved by
```swift
let complexRouteCharacters = CharacterSet(charactersIn: "*.:+?()[]\")

if pattern.rangeOfCharacter(from: complexRouteCharacters) != nil {
  return (nil, true, nil)
} 
```
Maybe we should measure which method is faster.

Sources/Kitura/RouterElement.swift
```diff
@@ -124,6 +135,44 @@ class RouterElement {
...
+    /// - Parameter response: the router response
+    /// - Parameter next: the closure for the next execution block
+    private func performSimpleMatch(path: String, request: RouterRequest, response: RouterResponse, next: @escaping () -> Void) {
+        let pathToMatch = path.isEmpty ? "/" : path
+        var matched: Bool
+        let matchedPath: String
+        
+        if allowPartialMatch {
+            matched = pathToMatch.hasPrefix(pattern!)
+            if matched && pattern! != "/" {
```
vadimeisenbergibm: I would add a `guard let pattern = pattern else` in the beginning of the function and would not use `pattern!`

Sources/Kitura/RouteRegex.swift
shmuelk: Fixed as we discussed.

Sources/Kitura/RouterElement.swift
shmuelk: Done.

Issue #1009, opened 2017-02-11
Response is not closed and hangs if router handlers do not call end() or next()


na-gupta:
Kitura Router.handle() calls `response.end()` in case none of the handlers call it, but this is dependent on the route handlers calling next(). Check and ensure RouterResponse end() has been invoked in it's deinitialization so we always end the response when we are done with it.

http://stackoverflow.com/questions/42176639/simplify-kitura-routerhandler-by-removing-next-closure/

youming-lin:
I would argue that it **should** be dependent on route handlers calling `next()`, especially for handlers that perform asynchronous work.

na-gupta:
That's a good point about handlers that do asynchronous work. I think we can introduce a flag for handlers to signal the need to defer ending the response, while making it more intuitive for the bulk of users. Are you aware of any uses of kitura where handlers are doing asynchronous work that requires deferring response.end() ?

youming-lin:
Anytime a handler needs to use asynchronous libraries, for example HTTP/database calls. I wrote an example as answer to an SO question: http://stackoverflow.com/a/40227985

na-gupta:
@youming-lin Makes sense. I think we can achieve the objective by checking and calling end() in RouterResponse's deinit unless I missed something else. Thanks!

youming-lin:
We need to discuss this change. IIRC in my tests the current Kitura behavior matches Node ExpressJS's; that is, if neither `next()` nor `response.end()` is called in the handler, then the request "hangs".

In the case that `next()` isn't called, do we 1) continue onto the next handler, or 2) end request processing here? If we insert `end()` in `RouterResponse`'s `denit` without making any other changes, 2) becomes the behavior, but I'm betting developers still expect 1) as the correct behavior.

na-gupta:
RouterResponse's denit will not be invoked until after `Router.handle(request:response:)` finishes and all handler/middleware objects with a reference to it exit that scope. Not sure why we would end request processing after the first handler does not call next()?

As long as any router/handlers/middleware have a strong reference to the `RouterResponse`, deinit should not be invoked.

youming-lin:
It's the same reason why not calling `next()` currently hangs the request. The middleware/element walking basically stops if `next()` isn't called. `Router` has the default response logic that *should* handle ending the response if it wasn't manually ended, but this code isn't executed if `next()` isn't called. This is because that code isn't reached due to the walker loop ending prematurely if `next()` isn't called.

[...]

PR #1010, opened 2017-02-11
Make sure Response is closed at deinit in case router handlers do not call end() or next()

na-gupta:
## Description
Kitura Router.handle() calls `response.end()` in case none of the handlers call it, but this is dependent on the route handlers calling next(). Check and ensure RouterResponse end() has been invoked in it's deinitialization so we always end the response when we are done with it.

## Motivation and Context
#1009 

## Checklist:

### DIFF
--- a/Sources/Kitura/RouterResponse.swift
+++ b/Sources/Kitura/RouterResponse.swift
@@ -121,6 +121,16 @@ public class RouterResponse {
         statusCode = .unknown
     }
 
+    deinit {
+        if !state.invokedEnd {
+            do {
+                try end()
+            } catch {
+                Log.warning("Error in RouterResponse end(): \(error)")
+            }
+        }
+    }
+
     /// End the response.
     ///
     /// - Throws: Socket.Error if an error occurred while writing to a socket.
--- a/Sources/Kitura/RouterResponse.swift
+++ b/Sources/Kitura/RouterResponse.swift
@@ -123,6 +123,10 @@ public class RouterResponse {
 
     deinit {
         if !state.invokedEnd {
+            if !state.invokedSend && statusCode == .unknown {
+                statusCode = .serviceUnavailable
+            }
+
             do {
                 try end()
             } catch {
--- a/Tests/KituraTests/TestResponse.swift
+++ b/Tests/KituraTests/TestResponse.swift
@@ -32,6 +32,7 @@ class TestResponse: KituraTest {
     static var allTests: [(String, (TestResponse) -> () throws -> Void)] {
         return [
             ("testSimpleResponse", testSimpleResponse),
+            ("testResponseNoEndOrNext", testResponseNoEndOrNext),
             ("testPostRequest", testPostRequest),
             ("testPostRequestTheHardWay", testPostRequestTheHardWay),
             ("testPostJSONRequest", testPostRequest),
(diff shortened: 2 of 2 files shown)


Tests/KituraTests/TestResponse.swift
```diff
@@ -81,6 +83,38 @@ class TestResponse: KituraTest {
...
+    func testEmptyHandler() {
+        performServerTest(router) { expectation in
+            self.performRequest("get", path:"/emptyHandler", callback: {response in
+                XCTAssertEqual(response?.statusCode, HTTPStatusCode.serviceUnavailable, "HTTP Status code was \(response?.statusCode)")
+                XCTAssertNotNil(response?.headers["Date"], "There was No Date header in the response")
+                do {
+                    let body = try response?.readString()
+                    XCTAssertNil(body)
+                } catch {
+                    XCTFail("No response body")
```
shmuelk: I think the error message should be "Error reading body", as there isn't suppose o be a body

Tests/KituraTests/TestResponse.swift
```diff
@@ -81,6 +83,38 @@ class TestResponse: KituraTest {
...
+    }
+
+    func testEmptyHandler() {
+        performServerTest(router) { expectation in
+            self.performRequest("get", path:"/emptyHandler", callback: {response in
+                XCTAssertEqual(response?.statusCode, HTTPStatusCode.serviceUnavailable, "HTTP Status code was \(response?.statusCode)")
+                XCTAssertNotNil(response?.headers["Date"], "There was No Date header in the response")
+                do {
+                    let body = try response?.readString()
+                    XCTAssertNil(body)
```
shmuelk: Please add a message to assert. Something like "A body wasn't expected, but was received \(body)"

Tests/KituraTests/TestResponse.swift
```diff
@@ -81,6 +83,38 @@ class TestResponse: KituraTest {
...
+    func testResponseNoEndOrNext() {
+        performServerTest(router) { expectation in
+            self.performRequest("get", path:"/noEndOrNext", callback: {response in
+                XCTAssertEqual(response?.statusCode, HTTPStatusCode.OK, "HTTP Status code was \(response?.statusCode)")
+                XCTAssertNotNil(response?.headers["Date"], "There was No Date header in the response")
+                do {
+                    let body = try response?.readString()
+                    XCTAssertEqual(body, "<!DOCTYPE html><html><body><b>noEndOrNext</b></body></html>\n\n")
+                } catch {
+                    XCTFail("No response body")
```
shmuelk: I think the error message should be "Error reading the body"

Sources/Kitura/RouterResponse.swift
```diff
@@ -121,6 +121,20 @@ public class RouterResponse {
         statusCode = .unknown
     }
 
+    deinit {
+        if !state.invokedEnd {
+            if !state.invokedSend && statusCode == .unknown {
```
shmuelk: I don't see a need to test again for !state.invokeSend. We are the deinit. No one is suppose to have a reference.

na-gupta: That is so if a handler has written to the response, but just didn't end the response, we send ok status, not service unavailable.

shmuelk: I didn't say to remove the second if. Just o remove the `!state.invokeEnd &&` part.

shmuelk: Never mind. I thought both if's were referencing state.invokeEnd.

shmuelk:
This can be merged once the build runs.

PR #1020, opened 2017-02-25
Add support for Range HTTP request header on static content

tcamin:
## Description

Adding support for Range HTTP header allows a client to selectively request specific chunks of a static content. Most modern browser support this header which is typically used to avoid downloading large contents all at once (i.e streaming videos).

## Motivation and Context

I've been using Kitura to build a service that serves video content and needed to be able to selectively change the current position of a video. Web browsers requires the server to support HTTP Range header.

## How Has This Been Tested?

Tests have been added including various combination of out of bounds data requests. A binary `streamdata` data file was added to verify that the requested bytes are correctly returned. The file, 100 bytes long, contains an ascending sequence of bytes from 0x00 to 0x63.

I've also verified on different browsers that the I was able to correctly change the current position of a video (feature that requires a correct implementation of the content range header).

## Checklist:


Sources/Kitura/staticFileServer/FileServer.swift
```diff
@@ -156,6 +156,85 @@ extension StaticFileServer {
...
+                
+                guard let rangeUnit = rangeInfo?.first,
+                    rangeUnit == "bytes",
+                    var byteRanges = rangeInfo?.last?.components(separatedBy: "-"),
+                    byteRanges.count == 2 else {
+                    response.statusCode = .badRequest
+                    return false
+                }
+                
+                guard let streamData = try? Data(contentsOf: URL(fileURLWithPath: filePath), options: [.mappedIfSafe]) else {
```
shmuelk: @tcamin I think you should open the file using FileHandle(forReadingAtPath: filePath), do a seek using FileHandle.seek(toFileOffset: ..) and then read in using FileHandle.readData(ofLength: end - start)

This is especially true if you're dealing with video files, which tend to be very very large.

tcamin: I can change that, however I thought that passing .mappedIfSafe was enough here. Will update the PR

shmuelk: I double checked and on the Linux port of Foundation (at 3.0.2) .mappedIfSafe is ignored. .alwaysMapped, seems to do what you want.

tcamin: Replaced . mappedIfSafe with .alwaysMapped and added a test to handle requests to empty files.

BTW I tried to replace the implementation using `FileHandler` however the linux tests failed (while on Mac they passed). Might there be an issue with it under Linux?

shmuelk:
Please try and test what happens if many clients simultaneously make requests. Does the "large" video file get mapped many times or only once?

tcamin:
Unfortunately the .alwaysMapped seems to be mapping the same file to different VM locations. I'll try to see if I understand what's not working with FileHandle under linux.

tcamin:
Apparently under Linux when getting file size by accessing `fileAttributes[FileAttributeKey.size]` one has to explicitly cast to NSNumber (and from there get the UInt64) instead of just casting directly to UInt64 as you do under MacOS. Seems to be working now.

kweinmeister:
@tcamin There are some merge conflicts now. Can you please review your PR again, so that we can consider merging it? Thank you!

tunniclm:
Hi @tcamin - thanks for your work on this PR. We've recently merged https://github.com/IBM-Swift/Kitura/pull/1161 to provide this functionality, so I think this PR no longer applies.

I'm going to close this PR for now but please feel free to reopen this or open a new issue/PR if you find anything missing or incorrect. Thanks again for your contribution.

PR #1026, opened 2017-03-03
Unify query string parsing for 'RouterRequest' and 'BodyParser'

naithar:
## Description
Corrects `URLEncodedBodyParser` dictionary parsing and provides correct representation of query parameter values for `.urlEncoded(...)` body.
Implementing `String` extension `. asUrlEncoded` make a single entry point for `url-encoded` parameters parsing.

## How Has This Been Tested?
I didn't add any new tests. 
Existing test that uses `.queryParameters` and `.urlEncoded(...)` body should cover this changes.

## Checklist:

### DIFF
--- a/Sources/Kitura/String+Extensions.swift
+++ b/Sources/Kitura/String+Extensions.swift
@@ -1,0 +1,43 @@
+/*
+ * Copyright IBM Corporation 2017
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


Sources/Kitura/String+Extensions.swift
```diff
@@ -0,0 +1,44 @@
...
+ * limitations under the License.
+ */
+
+import Foundation
+import LoggerAPI
+
+extension String {
+    
+    /// Parses percent encoded string into query parameters
+    var asUrlEncoded: [String : String] {
```
na-gupta: This is decoding, not encoding, and so should be called `urlDecoded`, or even better: `urlDecodedFieldValuePairs` to more clearly describe what it does.

Sources/Kitura/String+Extensions.swift
```diff
@@ -0,0 +1,44 @@
...
+        var result: [String:String] = [:]
+        
+        for item in self.components(separatedBy: "&") {
+            guard let range = item.range(of: "=") else {
+                continue
```
na-gupta: Needs `result[item] = nil` before `continue`

naithar:
Updated

PR #1035, opened 2017-03-16
Conditional Swift 3.1 support

carlbrown:
This will let this code compile and run (without warnings) on both swift 3.0.2 and 3.1 (March 15h snapshot).

We can remove this once we start requiring Swift 3.1, but this makes it easier for us to take our time to cut the latest tag once Xcode 8.3 is released.

We can take this or ignore it, but it might make our lives easier.

### DIFF
--- a/Sources/Kitura/RouterElement.swift
+++ b/Sources/Kitura/RouterElement.swift
@@ -30,7 +30,11 @@ class RouterElement {
 
     /// The regular expression
     #if os(Linux)
-        private var regex: RegularExpression?
+        #if swift(>=3.1)
+            private var regex: NSRegularExpression?
+        #else
+            private var regex: RegularExpression?
+        #endif
     #else
         private var regex: NSRegularExpression?
     #endif
--- a/Tests/KituraTests/TestRouteRegex.swift
+++ b/Tests/KituraTests/TestRouteRegex.swift
@@ -64,7 +64,11 @@ class TestRouteRegex: KituraTest {
 
     func testBuildRegexFromPattern() {
         #if os(Linux)
-            var regex: RegularExpression?
+            #if swift(>=3.1)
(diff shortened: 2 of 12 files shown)


na-gupta:
👍 LGTM

Sources/Kitura/RouterElement.swift
```diff
@@ -30,7 +30,11 @@ class RouterElement {
 
     /// The regular expression
     #if os(Linux)
-        private var regex: RegularExpression?
+        #if swift(>=3.1)
+            private var regex: NSRegularExpression?
+        #else
+            private var regex: RegularExpression?
+        #endif
```
youming-lin: Possible to use `RegularExpressionType ` here to eliminate need for compiler directives? The typealias should be accessible anywhere inside the package.

Tests/KituraTests/TestRouteRegex.swift
```diff
@@ -64,7 +64,11 @@ class TestRouteRegex: KituraTest {
 
     func testBuildRegexFromPattern() {
         #if os(Linux)
-            var regex: RegularExpression?
+            #if swift(>=3.1)
+                var regex: NSRegularExpression?
+            #else
+                var regex: RegularExpression?
+            #endif
```
youming-lin: Ditto here on using `RegularExpressionType`

Sources/Kitura/RouterElement.swift
carlbrown: I'm sure it would be. I wasn't trying to determine the intent of the original code, I was just following along with the existing pattern and trying to reduce the possibility of side-effects as much as possible..

youming-lin:
I'll go ahead and merge this. Further code cleanup can be done when we bump `.swift-version` for 3.1.

PR #1123, opened 2017-06-20
Swift 4 support and add travis builds

quanvo87:
The removal of `@escaping` on the parameter closures was required for Swift 4 because these were variadic closures, which are already escaping. The 3.1 compiler ignored this/was a bug.
https://bugs.swift.org/browse/SR-2444
https://lists.swift.org/pipermail/swift-users/Week-of-Mon-20160905/003185.html

In `RouterParameterWalker.swift`, the removal of `.filter` over the request parameters was required for Swift 4 because `filter`ing over a dictionary now correctly returns a dictionary in Swift 4. A `for` loop is used instead in order to build the required `[(String, String)]`.
https://stackoverflow.com/questions/32604897/swift-filter-dictionary

### DIFF
--- a/Sources/Kitura/RouterHTTPVerbs_generated.swift
+++ b/Sources/Kitura/RouterHTTPVerbs_generated.swift
@@ -28,7 +28,7 @@ extension Router {
     /// - Parameter handler: A comma delimited set of `RouterHandler`s that will be
     ///                     invoked when any request comes to the server.
     @discardableResult
-    public func all(_ path: String?=nil, handler: @escaping RouterHandler...) -> Router {
+    public func all(_ path: String?=nil, handler: RouterHandler...) -> Router {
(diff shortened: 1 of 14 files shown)


Sources/Kitura/RouterHTTPVerbs_generated.swift
```diff
@@ -28,7 +28,7 @@ extension Router {
     /// - Parameter handler: A comma delimited set of `RouterHandler`s that will be
```
shmuelk: The code in this file should *not* be modified. It is generated code, as it's name implies

Please fix the generator in `Scripts/generate_router_verbs.sh`, run the code and push the generated code

youming-lin:
Need to test changes again on latest snapshot due to Apple backing out of the tuple change: https://lists.swift.org/pipermail/swift-evolution-announce/2017-June/000386.html

PR #1148, opened 2017-09-08
Allow max Keep-Alive requests to be configurable

### LINKED ISSUE #1125: Allow tuning max requests per keepalive connection
I've been investigating Kitura's poor performance on OSX with https://github.com/tbrand/which_is_the_fastest

It appears that the benchmark client (written in crystal) doesn't perform well when having to establish new connections to the server.  Out of the 3 swift frameworks, looks like we're the only one that implements a limited number of requests per connection. Closing the connection every 100 requests really hurts the performance with this client.

Ideally I'd like to be able to increase the value (or disable the feature altogether). However, [the tunable in IncomingHTTPSocketProcessor.swift](https://github.com/IBM-Swift/Kitura-net/blob/1.7.12/Sources/KituraNet/HTTP/IncomingHTTPSocketProcessor.swift#L53) is private.

With the default (100), the benchmark reports:
```
bananas:which_is_the_fastest djones6$ ./bin/benchmarker swift
Language (Runtime)        Framework (Middleware)          Max [sec]       Min [sec]       Ave [sec]
------------------------- ------------------------- --------------- --------------- ---------------
swift                     vapor                            5.966822        5.813013        5.876588
swift                     perfect                          7.279527        7.109674        7.193408
swift                     kitura                          19.610924       18.156477       18.905314
```
But if I increase it to 100,000:
```
swift                     kitura                           6.989094        6.808682        6.898933
```
Currently, the only way I could 'fix' this would be some ugly hack in the makefile to edit the Kitura-net dependency before building.

djones6:
## Description
Enables the maximum number of subsequent Keep-Alive requests to be specified when creating an HTTP server.  A value of zero disables Keep-Alive.  A negative value allows unlimited requests, and is the default.

Resolves #1125 

Note that IBM-Swift/Kitura-net#216 is a prerequisite to this one and describes the motivation and context.

## How Has This Been Tested?

See IBM-Swift/Kitura-net#216

## Checklist:

### DIFF
--- a/Sources/Kitura/Kitura.swift
+++ b/Sources/Kitura/Kitura.swift
@@ -37,12 +37,17 @@ public class Kitura {
     /// - Parameter onPort: The port to listen on.
     /// - Parameter with: The `ServerDelegate` to use.
     /// - Parameter withSSL: The `sslConfig` to use.
+    /// - Parameter maxKeepAliveRequests: The number of additional requests that may be made per connection. A value of zero disables Keep-Alive. A negative value indicates that unlimited requests, and is the default.
     /// - Returns: The created `HTTPServer`.
     @discardableResult
-    public class func addHTTPServer(onPort port: Int, with delegate: ServerDelegate, withSSL sslConfig: SSLConfig?=nil) -> HTTPServer {
+    public class func addHTTPServer(onPort port: Int,
+                                    with delegate: ServerDelegate,
+                                    withSSL sslConfig: SSLConfig?=nil,
+                                    maxKeepAliveRequests maxRequests: Int = -1) -> HTTPServer {
         let server = HTTP.createServer()
         server.delegate = delegate
         server.sslConfig = sslConfig?.config
+        server.maxRequests = maxRequests
         httpServersAndPorts.append(server: server, port: port)
         return server
     }
--- a/Sources/Kitura/Kitura.swift
+++ b/Sources/Kitura/Kitura.swift
@@ -43,7 +43,7 @@ public class Kitura {
     public class func addHTTPServer(onPort port: Int,
                                     with delegate: ServerDelegate,
                                     withSSL sslConfig: SSLConfig?=nil,
-                                    keepAlive keepAliveState: HTTPServerResponse.KeepAliveState = .unlimited) -> HTTPServer {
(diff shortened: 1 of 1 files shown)


DonaldHays:
Hi!

I am neither much of a user of nor contributor to Kitura, but I was looking at the project and stumbled upon this pull request and had some feedback that I thought may be of value. But since I haven't spent much time with the project, I readily acknowledge that I may be totally naïve to best practices, standards, or precedent for this project that may totally invalidate what I would like to say. I am offering this feedback from only the perspective of a general Swift developer.

This pull request (and the one it depends upon) introduces the concept of a configurable number of requests to be permitted per Keep-Alive connection. There are three ways it can be configured: with Keep-Alive disabled, Keep-Alive enabled and allowing a finite number of requests, or Keep-Alive enabled and allowing an unbounded number of requests. The configuration is made with an `Int`. A sentinel value (negative numbers) is used to indicate unlimited requests. A value of 0 is _sort of_ a sentinel value in how it's used to disable Keep-Alive entirely, though it's also kind of intuitively reasonable when one thinks of what 0 implies.

Swift in general seems to try to move away from sentinel values—as can be seen in APIs like `index(of:)`, which returns `nil` instead of a sentinel value if an element is not found. I feel that `-1` indicating "unlimited" is a bit atypical for Swift API design, and so I offer two alternatives, and would like your feedback if you think either of them is reasonable.

If either of these alternatives are adopted, they would also need to be expressed in the parent pull request.

# Alternative 1: An Optional

In this design `maxRequests` is of type `Int?`, where `nil` indicates "no limit", and a non-nil value indicates "this is the maximum permitted", with 0 indicating that Keep-Alive should be disabled.

# Alternative 2: An Enum

In this design `maxRequests` is not an `Int` at all, and is rather an enum with two or three cases, e.g.:
```swift
enum KeepAliveMode {
  case disabled
  case enabled(maximumNumberOfRequests: Int?)
}
```
or
```swift
enum KeepAliveMode {
  case disabled
  case unlimited
  case limited(maximumNumberOfRequests: Int)
}
```

djones6:
@youming-lin This should be ready to go once IBM-Swift/Kitura-net#216 has been tagged.

PR #1161, opened 2017-10-18
Implement Range Requests

nacho4d:
Implementation of [Range Requests - RFC7233](https://tools.ietf.org/html/rfc7233).
Please review the code. I am happy to fix/refactor anything here :)

## Description

FileServer now recognizes Range requests (headers like: `Range: bytes=0-9`) and properly returns 206 (Partial Content) responses as per RFC spec

## Motivation and Context
I was using Kitura for a simple static server and found out that  `<video>` tags do not play in Safari. Safari do this Range requests before streaming videos. Now it works in Safari too!
(Chrome and Firefox already worked)

## How Has This Been Tested?
* Several tests have been added
* I have run the tests in macos

### Long description:
- Added(ported) a RangeHeader and a parser based on nodejs [range-parser](https://github.com/jshttp/range-parser)
- Tests for the range parser are also ported
- Added an option to disable Range request (As same as nodejs)
- Added an enum to MultiPartBodyParser to be able to read complex 206 responses 
### Things I have not done
- ~~Ranges do not get merged. Ex: `bytes=0-9,10-20` would become `0-20` since these are adyacent and or overlapping. (As per RFC this is optional, this is not complicated, just need to port more code from the range parser)~~ *I have implemented this in the last commit*. 
- `If-Range` headers are still ignored. (RFC says if If-Range header is present and if file has changed after the condition then server should return 200 instead of 206. This requires more reading of the spec but doable in some days I guess)

## Checklist:


nacho4d:
## CodeClimate complaining:
I added a TODO in tests. Because I cannot do HEAD requests when using ClientRequest. Response is always nil a test won't work.  Should I remove the TODOs for now?

## Travis complaining:
I have addressed various build issues specially for swift 3.x and Linux compiler however Travis is still complaining. 
Weird, all tests succeed in macos (swift4/xcode9) but fail in Travis with swift-4.0-DEVELOPMENT-SNAPSHOT-2017-09-11-a I will look into this.

nacho4d:
I have removed the TODO to make codeclimate happy.
Source is compatible with old 3.1 swift so all tests are passing now. I don't know why but travis is not running on macos? or do I just need to wait more? (All tests pass in my local mac machine)

nacho4d:
Can someone give a him why the macos build is failing? All tests are ok but some command after it is failing, is is because of changes here?

nacho4d:
After a week, this is ready for review :)

nacho4d:
I resolved conflicts risen due to recent changes in master branch

tunniclm:
This looks very good.

Just wanted to add one note to the list of implementation notes above:

* This implementation never produces "Range Not Satisfiable" and instead uses the approach noted in Section 4.4 (page 15) of RFC7233 ("Note: Because servers are free to ignore Range, many implementations will simply respond with the entire selected representation in a 200 (OK) response. [...]") when encountering requests with invalid or unsatisfiable ranges

I have a few final questions:

* @nacho4d : The disabled test should work with the latest release of Kitura-net (`1.7.19`) 
* @seabaylea: Do we need to update a `NOTICES` file or make reference to the MIT license/copyright on node-parser for the code in `RangeHeader.swift`?

Thanks very much for your work on this @nacho4d!

rolivieri:
@nacho4d Just echoing tunniclm comments above, thanks for this PR and including a very good set of test cases to validate the new functionality.

nacho4d:
It got merged!! yeeeah!
Thank you @rolivieri @seabaylea @tunniclm for the time spend reviewing this. I will address the disabled test, etc  in another PR.

PR #1162, opened 2017-10-19
Expose document root dir used by StaticFileServer

tunniclm:
## Description
This change exposes the document root / base directory used by the StaticFileServer middleware.

## Motivation and Context
The absolute path used for the document root is based on the code in `StaticFileServer.ResourcePathHandler.getAbsolutePath()` and can vary depending on the environment.
In order to work out where the directory is a user would have to reimplement the logic of this non-public code which is inconvenient and brittle to changes in the internals of Kitura.

Instead this PR adds a new property to the StaticFileServer API to provide access to the computed directory.

## How Has This Been Tested?
Tested in an example application and with a simple test case added to `KituraTests/TestStaticFileServer.swift`.

## Checklist:

### DIFF
--- a/Sources/Kitura/staticFileServer/StaticFileServer.swift
+++ b/Sources/Kitura/staticFileServer/StaticFileServer.swift
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corporation 2016
+ * Copyright IBM Corporation 2016,2017
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


tunniclm:
Feel free to suggest an alternative name to `absoluteRootPath`. This name was intending to evoke "document root" as used in web servers like Apache HTTP Server.

seabaylea:
Could you add Jazzy doc comments for the new field?

PR #1177, opened 2017-11-14
Migrate to swift 4.0.2

ShihabMehboob:


.travis.yml
```diff
@@ -25,9 +25,8 @@ matrix:
       dist: trusty
       sudo: required
       services: docker
-      env: DOCKER_IMAGE=ubuntu:16.04 SWIFT_SNAPSHOT=4.0
```
sandmman: Could you put the DOCKER_IMAGE env variable back? We still need to test 16.04.

sandmman: Could you put the DOCKER_IMAGE env variable back? We still need to test 16.04.

rolivieri: @sandmman @ShihabMehboob made the requested change. The travis builds finally completed. I will process the PR now, thanks!

PR #1187, opened 2017-11-20
Finalizing Range Requests implementation

nacho4d:
Implementation of 416 (Not Satisfiable Range) responses and honor `If-Range` headers when available in Range requests.
Please review this code, I am happy to fix anything here :]

## Description

- Extended RangeHeader to throw errors so FileServer is able to decide between 206 and 416 responses.
- Implemented Not Satisfiable Range responses (416), part of [RFC7233 section 4.4](https://tools.ietf.org/html/rfc7233#section-4.4)
- Honor `If-Range` header (when given) as per [RFC7233 section 3.2](https://tools.ietf.org/html/rfc7233#section-3.2)
    - When it is of the ETag form, it must match the ETag for the file at its current state
    - When it is of the Last-Modified form, it must be the same as the last modified date of the file

> Informally, `If-Range` meaning is "if the entity is unchanged, send me the part(s) that I am missing; otherwise, send me the entire new entity".

## Motivation and Context
It completes implementation of Range Requests (RFC7233) :) which was previously implemented with minimun basic functionality in #1161 

## How Has This Been Tested?
- Added various tests to check all functionality

## Checklist:


nacho4d:
Ready for review 😀

nacho4d:
thank you @rolivieri  :)

PR #1189, opened 2017-12-06
Issue.488 - Updating Kitura Docs

Andrew-Lees11:
## Description
Updated the Jazzy Docs for Kitura including a usage example for each class. This involves no changes to the code since it is only changing comments. The new Docs require review to ensure they properly describe the classes as well as general wording and grammar checks.


helenmasters:
Andy, please can you fix the fact that the contributing guidelines link in the README.md file doesn't work from the context of the API reference site (it does work when used directly from the README.md file on github). I realise that this isn't one of the files you've changed but to my mind it forms part of this documentation work so should be fixed as part of this changeset.

Sources/Kitura/bodyParser/BodyParser.swift
```diff
@@ -21,11 +21,13 @@ import Foundation
 
 // MARK: BodyParser
 
-/// The BodyParser parses the body of the request prior to sending it to the handler. It reads the Content-Type of the message header and populates the `RouterRequest` body field with a `ParsedBody` enumeration (e.g. json, raw, text, urlEncoded)
-/// In order for the BodyParser to be used it must first be registered with any routes that are interested in the ParsedBody payload. The example below registers all routes to the BodyParser middleware.
+/// The `BodyParser` parses the body of the request prior to sending it to the handler. It reads the Content-Type of the message header and populates the `RouterRequest` body field with a `ParsedBody` enumeration (e.g. json, raw, text, urlEncoded)
+/// In order for the BodyParser to be used it must first be registered with any routes that are interested in the ParsedBody payload.
```
helenmasters: Missing a full stop at the end of line 24.

Sources/Kitura/bodyParser/BodyParser.swift
```diff
@@ -115,9 +117,9 @@ public class BodyParser: RouterMiddleware {
 
     /// Read incoming message for Parse
     ///
-    /// - Parameter message: message coming from the socket
+    /// - Parameter message: Message coming from the socket
     /// - Parameter parser: ((NSData) -> ParsedBody?) store at parserMap
-    /// - Returns: the parsed body
+    /// - Returns: The parsed body
```
helenmasters: Why are there no usage examples for the init, parse, readBodyData and handle calls?

Sources/Kitura/bodyParser/BodyParser.swift
```diff
@@ -131,9 +133,9 @@ public class BodyParser: RouterMiddleware {
 
     /// Read the body data of the request.
     ///
-    /// - Parameter with: the socket reader.
+    /// - Parameter with: The socket reader.
     /// - Throws: Socket.Error if an error occurred while reading from a socket.
-    /// - Returns: data for the body.
+    /// - Returns: Data for the body.
```
helenmasters: "Data for the body." doesn't make sense to me how about something like "The body data of the request."?

Sources/Kitura/contentType/ContentType.swift
```diff
@@ -19,30 +19,37 @@ import LoggerAPI
 
 // MARK: ContentType
 
-/// A set of APIs to work with Content-Type headers, whether to generate the value
-/// or to determine if it's an acceptable value.
+/**
+The `ContentType` class provides functions to determine the MIME content type for a given file extension. The user can pass in a complete file name e.g. "foo.png" or just the file extension e.g. "png", or they can pass in both a MIME content type and a file extension and query whether the they match.
```
helenmasters: Remove "the" from "the they match"

Sources/Kitura/Kitura.swift
```diff
@@ -22,7 +22,20 @@ import Dispatch
...
+```swift
+let router = Router()
+let port: Int = 8080
+public func run() throws{
+    ...
+    Kitura.addHTTPServer(onPort: port, with: router)
+    Kitura.run()
+}
+```
+In this example, we create a run function inside a server Application.swift file, which will create a Kitura server on the specified port, using the given router and then run the created server.
```
helenmasters: The example text needs to go above the code snippet.

helenmasters: @tunniclm or @rolivieri -  I notice that we are saying two different things in this file - in the example code we call Kitura run() function inside a server Application.swift file, however, within the run() function we have a note: This function never returns - it should be the last call in your main.swift file.
Where is the recommended place to put the Kitura run() call as I'd like our usage example to reflect this (or definitely to have some better wording as to why either location may be suitable)?

helenmasters: The following sentences need full stops at the end - "Stop all registered servers", "Start all registered servers and return"

helenmasters: /// - Parameter unregister: If servers should be unregistered after stopped (default true).

Please add the words - they are - so it reads unregistered after they are stopped.

ianpartridge:
@Andrew-Lees11 this is stale - what's the status?

djones6:
Closing after discussion: this is being split into smaller updates.

ianpartridge:
Do we need to keep the branch around?

Issue #1188, opened 2017-11-23
Content-Type header missing with codable routing


djones6:
When invoking a route that uses codable routing, JSON data is returned in the message body, but the `Content-Type` header is missing.

Example project:  https://github.com/djones6/CodableTest

Here's the route and the response:
```swift
    router.get("/codable") {
        (completion: ([Hello]?, RequestError?) -> Void) in
        let hello = Hello(message: "Hello World")
        completion([hello], nil)
    }
```
```
$ curl -i http://localhost:8080/codable
HTTP/1.1 200 OK
Date: Thu, 23 Nov 2017 15:53:40 GMT
Content-Length: 27
Connection: Keep-Alive
Keep-Alive: timeout=60

[{"message":"Hello World"}]
```
The equivalent using the existing API does include the header:
```swift
    router.get("/json") {
        request, response, _ in
        let hello = Hello(message: "Hello World")
        try response.send(json: [hello]).end()
    }
```
```
$ curl -i http://localhost:8080/json
HTTP/1.1 200 OK
Date: Thu, 23 Nov 2017 15:53:48 GMT
Content-Type: application/json
Content-Length: 27
Connection: Keep-Alive
Keep-Alive: timeout=60

[{"message":"Hello World"}]
```

seabaylea:
This has just been reported by @svanimpe as well.

In each of the "safely" functions we do the encoding directly, e.g.  for `postSafely()`:
https://github.com/IBM-Swift/Kitura/blob/master/Sources/Kitura/CodableRouter.swift#L235
```swift
    let encoded = try JSONEncoder().encode(result)
    response.status(.created)
    response.send(data: encoded)
```
we have two choices:
1) Add a call to `response.headers.setType("json")`
2) Call `response.send(result)` instead and delegate the encoding to the send call

Approach #2 makes the most sense as there's no duplication of code/function. Currently there's no way to tell the response which encoder to use - that's not a problem at the moment, but we'll need to provide a setter in the future once we support other encoding types.

tunniclm:
@ddunn2 I think this issue is solved by https://github.com/IBM-Swift/Kitura/pull/1190

ddunn2:
@tunniclm I'll have a read through of the item you've linked, thanks! :)

rolivieri:
@ddunn2 @tunniclm Yes, @sandmman noticed this while implementing the two new proposals and has added code to address this issue.

christiancompton:
[...]

PR #1190, opened 2017-12-08
Kitura Evolution #0002 & #0003

sandmman:
This provides the implementation of Kitura Evolution Proposals [#0002 - Query Parameters Enhancement](https://github.com/IBM-Swift/evolution/blob/master/Kitura/Proposals/0002-query-parameters-enhancement.md) and [#0003 - Codable Query Parameters](https://github.com/IBM-Swift/evolution/blob/master/Kitura/Proposals/0003-codable-query-parameters.md).

## Motivation and Context
These proposals aim to simplify the developer experience by reducing boilerplate code in [#0002](https://github.com/IBM-Swift/evolution/blob/master/Kitura/Proposals/0002-query-parameters-enhancement.md) and providing type-safe query parameter handling as part of the codable router in [#0003](https://github.com/IBM-Swift/evolution/blob/master/Kitura/Proposals/0003-codable-query-parameters.md).

#### Dependencies
This PR is dependent upon KituraContracts [PR #8](https://github.com/IBM-Swift/KituraContracts/pull/8), which adds the Query protocol in order to eliminate potential ambiguity between the delete by identifier and the new delete by query method signatures. 

```swift
/// Current
func delete<Id: Identifier>(_ route: String, handler: @escaping (Id, @escaping ResultClosure) -> Void)

/// New
func delete<Q: Query>(_ route: String, handler: @escaping (Q, @escaping ResultClosure) -> Void)
```

**NOTE:** If accepted,  the package.swift will have to be reverted back to use the updated KituraContracts code/version.

#### Additional Bug Fix
None of our codable routes were returning `Content-Type=application/json`. I've updated their handlers accordingly. In doing so, this modified the [send codable](https://github.com/IBM-Swift/Kitura/blob/ade1c270aab30d7570d97884714fe43c9ac68d4d/Sources/Kitura/RouterResponse.swift#L449) RouterRequest method to handle the error instead of simply logging it. We may actually want to throw here or allow the user to pass in their own status variables for success and for failure. Thoughts?

## How Has This Been Tested?
Unit Tests

## Checklist:
- [ x] If applicable, I have updated the documentation accordingly.


tunniclm:
I like these changes, they are very similar (in the use of `response.send()` instead of `JSONEncoder()`) to some I have drafted locally. One thing I also hit is the way `send()` currently handles errors, which I don't like and I was glad to see mentioned in the changes.

My initial thought on that would be that Kitura's top-level `catch` ought to default to returning `.internalServerError` (and logging the failure) for a handler that throws, that way any true "bail out all is lost" exceptions can just be left uncaught by the handler which nicely simplifies the API. I'm not sure whether there are any gotchas in making sure these exceptions get properly transmitted up the callback chain.

With that, we could just make all the `send()` implementations that could fail throw an exception. This gives the caller the choice to handle it or bail out on the response handling (either explicitly by catching, or implicitly by allowing to exception to bubble up).

tunniclm:
I think this solves https://github.com/IBM-Swift/Kitura/issues/1188 also

rolivieri:
The following are the changes @sandmman will make to this PR before we can merge it. If anyone in the team has a different view, let us know so we can discuss. Otherwise, we will move forward.

1) Regarding error handling in the [`send()`](https://github.com/IBM-Swift/Kitura/blob/ade1c270aab30d7570d97884714fe43c9ac68d4d/Sources/Kitura/RouterResponse.swift#L449) function, we noticed that we are not being consistent in the API provided in the `RouterResponse` object. There are other functions in that same object that do throw an exception when an error occurs instead of catching the exception and just logging an error. For instance, see [`send()`](https://github.com/IBM-Swift/Kitura/blob/ade1c270aab30d7570d97884714fe43c9ac68d4d/Sources/Kitura/RouterResponse.swift#L391) and [`end()`](https://github.com/IBM-Swift/Kitura/blob/ade1c270aab30d7570d97884714fe43c9ac68d4d/Sources/Kitura/RouterResponse.swift#L152).

    Therefore, our plan is to change the method signature of the `send()` methods so they al throw an exception instead of swallowing the error. This will make the API consistent across the `RouterResponse` object. The following are the `send()` methods that will be affected by this change:

    - [`send()`](https://github.com/IBM-Swift/Kitura/blob/ade1c270aab30d7570d97884714fe43c9ac68d4d/Sources/Kitura/RouterResponse.swift#L269)
    - [`send()`](https://github.com/IBM-Swift/Kitura/blob/ade1c270aab30d7570d97884714fe43c9ac68d4d/Sources/Kitura/RouterResponse.swift#L290)
    - [`send()`](https://github.com/IBM-Swift/Kitura/blob/ade1c270aab30d7570d97884714fe43c9ac68d4d/Sources/Kitura/RouterResponse.swift#L449)
    - [`send()`](https://github.com/IBM-Swift/Kitura/blob/ade1c270aab30d7570d97884714fe43c9ac68d4d/Sources/Kitura/RouterResponse.swift#L474)

2) Developers using any of the `send()` methods described above can then expect the same behavior… and if they are calling any of these methods from the main route handler function, they can let the exception bubble up all the way to the framework if they wish to do so (the current signature of the route handler allows throwing exceptions). In such cases, the framework will catch the exception and return a 404 error (note that sending 404 is the current behavior... we can discuss if 404 was the right choice... but I would say that's a separate discussion beyond the scope of this PR). In the case they are calling any of these methods from another method (e.g. say asynchronously, like a callback function that processes the results from a db invocation), then they are responsible for handling the exception.

3) Note that none of the `send()` methods described above in point 1 are available to the developer when using any of the Codable Routing APIs (the response object is not exposed in the API in such cases). Also, note that the signature for the handler methods associated with the Codable routing APIs currently does not allow throwing exceptions. We prefer to keep that signature the way it is. Hence, our plan is to catch the exception thrown by the call to the `send()` method, log it as an error, and return internal server error to the client. For instance, see: https://github.com/IBM-Swift/Kitura/blob/ade1c270aab30d7570d97884714fe43c9ac68d4d/Sources/Kitura/CodableRouter.swift#L275 This is consistent with the current experience we currently provide to developers leveraging any of the Codable APIs.

iDevPro:
I was added route:

```router.get("/api/player/all") { (respondWith: ([Player]?, RequestError?) -> Void) in

            let players = self.data.getPlayers().sorted(by: >)

           respondWith(players, nil)
        }```

But got error:
`Cannot invoke 'get' with an argument list of type '(String, @escaping (([Player]?, Server.RequestError?) -> Void) -> ())'`

Issue #1200, opened 2017-12-19
Build errors with Kitura 2.1.2


sjanuary:
I'm seeing build errors building SwiftMetrics against Kitura 2.1.2.  I don't see these with 2.1.0.  Running Swift 4.0.0 RELEASE on Linux.
```
Fetching https://github.com/IBM-Swift/Kitura.git
Fetching https://github.com/IBM-Swift/Kitura-WebSocket.git
Fetching https://github.com/IBM-Swift/SwiftyRequest.git
Fetching https://github.com/IBM-Swift/Swift-cfenv.git
Fetching https://github.com/IBM-Swift/SwiftyJSON.git
Fetching https://github.com/RuntimeTools/omr-agentcore
Fetching https://github.com/IBM-Swift/Kitura-net.git
Fetching https://github.com/IBM-Swift/Kitura-TemplateEngine.git
Fetching https://github.com/IBM-Swift/KituraContracts.git
Fetching https://github.com/IBM-Swift/LoggerAPI.git
Fetching https://github.com/IBM-Swift/BlueSocket.git
Fetching https://github.com/IBM-Swift/CCurl.git
Fetching https://github.com/IBM-Swift/BlueSSLService.git
Fetching https://github.com/IBM-Swift/CEpoll.git
Fetching https://github.com/IBM-Swift/BlueSignals.git
Fetching https://github.com/IBM-Swift/OpenSSL.git
Fetching https://github.com/IBM-Swift/BlueCryptor.git
Fetching https://github.com/IBM-Swift/CircuitBreaker.git
Fetching https://github.com/IBM-Swift/Configuration.git
Fetching https://github.com/IBM-Swift/FileKit.git
Cloning https://github.com/IBM-Swift/Swift-cfenv.git
Resolving https://github.com/IBM-Swift/Swift-cfenv.git at 6.0.0
Cloning https://github.com/IBM-Swift/BlueSocket.git
Resolving https://github.com/IBM-Swift/BlueSocket.git at 0.12.79
Cloning https://github.com/IBM-Swift/LoggerAPI.git
Resolving https://github.com/IBM-Swift/LoggerAPI.git at 1.7.1
Cloning https://github.com/IBM-Swift/BlueCryptor.git
Resolving https://github.com/IBM-Swift/BlueCryptor.git at 0.8.21
Cloning https://github.com/IBM-Swift/CircuitBreaker.git
[...]

PR #1201, opened 2017-12-19
Revert and redo warning fix

tunniclm:
Rollback the changes in 2.1.2 which break on Swift 4.0.0 and reapply them in a subsequent commit.

## Description
`2.1.2` fixes some compile warnings on Linux on Swift 4.0.2/4.0.3 but breaks builds on Swift 4.0.0 (as it takes advantage of a fix in corelibs-foundation that is not present in Swift 4.0.0).

This PR rolls back that change to be tagged as `2.1.3` and reapplies it to be tagged as `2.2.0` (since it is breaking). This could probably have been done by tagging 2.1.3 to the same commit as 2.1.1 and tagging 2.2.0 as the same as 2.1.2, however, that could be really confusing so I think it is better to revert and redo the change (simple to do with `git revert`).

## Motivation and Context
This change prevents compile errors on Swift 4 in a patch update. Instead we will make it a minor update (which Kitura is using for breaking changes). Should resolve https://github.com/IBM-Swift/Kitura/issues/1200.

## How Has This Been Tested?
Using `swift test`.

## Checklist:

### DIFF
--- a/Sources/Kitura/staticFileServer/CacheRelatedHeadersSetter.swift
+++ b/Sources/Kitura/staticFileServer/CacheRelatedHeadersSetter.swift
@@ -70,13 +70,8 @@ extension StaticFileServer {
                 let size = fileAttributes[FileAttributeKey.size] as? NSNumber else {
                 return nil
             }
-            #if !os(Linux)
-                // https://bugs.swift.org/browse/SR-5850
-                let sizeHex = String(Int(truncating: size), radix: 16, uppercase: false)
-            #else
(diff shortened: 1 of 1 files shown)


tunniclm:
Alternative to tagging 2.2.0 here we could make the warning fix apply only to Swift 4.0.2 and greater, then we could do just that in 2.1.3.

tunniclm:
I'm going to close this in favour of https://github.com/IBM-Swift/Kitura/pull/1202

Issue #1207, opened 2018-01-16
KituraNet needed to be included to send custom response status even Kitura depends on Kitura-Net


CosmicYogi:
I was working on my project today and everything was working fine till a Saw a weird behavior.
I wanted to send custom `response codes` for few situations so I refactored the code and started to work on it.

There are 2 issues I faced -
1. You cannot send custom code without importing kituraNet and when you `Option + click` on it to get the info about it, It says declared in Kitura and nothing else I mean either the documentation can be updated or the situation can be better managed it becomes very confusing if someone doesn't know about it.

2. There is an ambiguity faced when you import `KituraNet` in your project and try to catch an error as `Error`.

tunniclm:
I'm not sure what you mean by custom status code, can you past an example?

It is true that `HTTPStatusCode` is defined in the `KituraNet` module and you must import the defining module to refer to a type directly. Often you can avoid this by not naming type and relying on type inference. (Example: `response.status(.OK)` does not require `KituraNet` to be imported but `response.status(HTTPStatusCode.OK)` does)

tunniclm:
It is unfortunate that `KituraNet` defines an `Error` type, and this conflicts with the one in `Swift`.
We should consider renaming it in a future major update, since it forces users of `KituraNet` to disambiguate the types with their module name (`KituraNet.Error` / `Swift.Error`) to use them.

CosmicYogi:
Yeah, doing this the way `response.status(.OK)` worked without importing `KituraNet`. Thanks for the clarification.
It's good that it's working but I think it should be more clear I mean either it should be mentioned in the documentation so that programmers don't get confused while using it.
I also have a doubt I will PM you on Slack regarding it.

PR #1208, opened 2018-01-16
Added documentation comment related to HTTPStatusCode

CosmicYogi:
Added a documentation comment so that the confusion related to importing `KituraNet` can be removed.

## Description

I had added documentation comment in `RouterResponse.swift` related to function `status` which takes `HTTPStatusCode` as an input parameter but you need to import `KituraNet` to actually pass that parameter because it's described there.
Although `Kitura` depends on `KituraNet` but then also it doesn't work as expected it will give you an error if you will try to do `response.status(HTTPStatusCode.unauthorized)` without importing `KituraNet` and will also give an error if you import `Kitura`.

## Motivation and Context

This change will remove confusion from the developer's mind related to import of `KituraNet` I mean this might not be the perfect or very robust solution but will definitely help till there is no robust solution.
https://github.com/IBM-Swift/Kitura/issues/1207

## How Has This Been Tested?

As the changes are in the form of comments so they are not affecting the logic so the code coverage is same.

## Checklist:

### DIFF
--- a/Sources/Kitura/RouterResponse.swift
+++ b/Sources/Kitura/RouterResponse.swift
@@ -313,6 +313,7 @@ public class RouterResponse {
     ///
     /// - Parameter status: the HTTP status code.
     /// - Returns: this RouterResponse.
+    /// - Note: *KituraNet* needed to be imported to to set HTTPStatusCode. Can be done in way `response.status(HTTPStatusCode.unauthorized)`
     public func send(status: HTTPStatusCode) -> RouterResponse {
         guard !state.invokedEnd else {
             Log.warning("RouterResponse send(status:) invoked after end() for \(self.request.urlURL)")


tunniclm:
I've been playing with the idea of bridging `HTTPStatusCode` to Kitura with a typealias so that importing KituraNet is not required. This would mean you shouldn't hit the problem in the first place. I'll put something together for people to look at.

djones6:
@CosmicYogi #1217 has been merged, so this update is no longer applicable.  Please could you verify your issue is resolved?

ianpartridge:
I'm closing this as we now bridge `HTTPStatusCode` via a typealias. Thanks for the PR!

CosmicYogi:
Great.
Sorry, I was not available for last few weeks so the conversation was delayed from my side.
Yep, Close it.

PR #1214, opened 2018-01-29
Add RequestError body codable

tunniclm:
Add the capability to provide a `Codable` object to `RequestError` that will be encoded and sent by the `Router` as the response body when passed to the completion handler of a codable route.

## Description
This PR is one part of a pair, this one is for `Kitura`, the other is for `KituraContracts` (https://github.com/IBM-Swift/KituraContracts/pull/12).

**In the `KituraContracts` PR** we add a new generic constructor to `RequestError` that can specify a "base" request error and a body `Codable`, for example: `RequestError(.unavailable, body: health.status)`.

This will assign the object to a new stored property `body` as a (type-erased) `Codable`. The fact the type is erased allow the `RequestError` struct to remain non-generic (changing it to be generic would be a breaking change to the API).

However, this means that the real type of the `body` is no longer know. In order for the `Router` to encode the `body` into `Data` in some format (currently we only support JSON, but in the future there may be other formats) we need to be able to access the real type to pass to the encoder.

To achieve this we also store a closure in a second new stored property `bodyData` at the same time (from within the generic `init` where we still have access to the real type of the body codable). This closure accepts an argument describing the format into which you want to encode the body and returns a `Data` containing the encoded form of the body.

A new struct `BodyFormat` has been added to enumerate the different supported format. Currently only JSON is supported. This was made as a struct rather than an enum to allow new formats to be added without breaking the API (enums are exhaustive and adding a case would break any switch statements relying on that.)

The `bodyData` closure performs a switch on the `BodyFormat` provided and calls the appropriate encoder on the body to convert it to `Data`. If the `BodyFormat` is not recognised, then an `UnsupportedBodyFormatError` (another new type added in this change) will be thrown. Since  `BodyFormat` only has a private initializer, no other code should be able to create them, and provided the generic initializer of `RequestError` and `BodyFormat` are kept in sync, then `UnsupportedBodyFormatError` should never be thrown.

**In this PR** the `Router` has been updated so that codable routes will check for the existence of a `bodyData` in any `RequestError` returned by a codable route handler, and if it is non-nil then will use it to get a `Data` to send as the body of the response.

This PR also adds a number of tests, and to prevent excessive code duplication a new test builder API has been added to allow construction of concise tests in a declarative style.

## Motivation and Context
This change allows codable route handlers to customise the body of error responses. One example of where this feature is needed is the health route generated by the Swift Server Generator and kitura-cli.

This route should send `HTTP 200 OK` when the application is healthy and `HTTP 503 Service Unavailable` when the application is not working. It should also provide a response body that is a JSON object describing the status of the application and each dependent service. This body should be sent both when the application is healthy and when it is not, and is arguably more important in the latter case as it will give information about which service is not available.

Without a custom body for `RequestError` it is not possible to send a response body if the status code is not 2xx (usually HTTP 200 OK, but it varies depending on the HTTP method, eg: `delete` uses HTTP 204 No content).

## How Has This Been Tested?
New tests have been added, tested with `swift test`.

## Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -31,7 +31,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/IBM-Swift/Kitura-net.git", .upToNextMinor(from: "2.0.0")),
         .package(url: "https://github.com/IBM-Swift/Kitura-TemplateEngine.git", .upToNextMinor(from: "1.7.0")),
-        .package(url: "https://github.com/IBM-Swift/KituraContracts.git", .branch("add_requesterror_body"))
+        .package(url: "https://github.com/IBM-Swift/KituraContracts.git", .upToNextMinor(from: "0.0.19")),
     ],
     targets: [
         // Targets are the basic building blocks of a package. A target can define a module or a test suite.
--- a/Package.swift
+++ b/Package.swift
@@ -31,7 +31,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/IBM-Swift/Kitura-net.git", .upToNextMinor(from: "2.0.0")),
         .package(url: "https://github.com/IBM-Swift/Kitura-TemplateEngine.git", .upToNextMinor(from: "1.7.0")),
-        .package(url: "https://github.com/IBM-Swift/KituraContracts.git", .upToNextMinor(from: "0.0.17"))
+        .package(url: "https://github.com/IBM-Swift/KituraContracts.git", .branch("add_requesterror_body"))
     ],
     targets: [
         // Targets are the basic building blocks of a package. A target can define a module or a test suite.
--- a/Sources/Kitura/CodableRouter.swift
+++ b/Sources/Kitura/CodableRouter.swift
@@ -273,6 +273,7 @@ extension Router {
                             let status = self.httpStatusCode(from: err)
                             response.status(status)
                             if let bodyData = err.bodyData {
+                                response.headers.setType("json")
                                 response.send(data: try bodyData(.json))
                             }
                         } else {
@@ -326,6 +327,7 @@ extension Router {
                             let status = self.httpStatusCode(from: err)
                             response.status(status)
                             if let bodyData = err.bodyData {
+                                response.headers.setType("json")
                                 response.send(data: try bodyData(.json))
                             }
                         } else {
@@ -387,6 +389,7 @@ extension Router {
                             let status = self.httpStatusCode(from: err)
                             response.status(status)
                             if let bodyData = err.bodyData {
+                                response.headers.setType("json")
                                 response.send(data: try bodyData(.json))
                             }
                         } else {
@@ -441,6 +444,7 @@ extension Router {
                             let status = self.httpStatusCode(from: err)
                             response.status(status)
                             if let bodyData = err.bodyData {
+                                response.headers.setType("json")
                                 response.send(data: try bodyData(.json))
                             }
                         } else {
@@ -476,6 +480,7 @@ extension Router {
                         let status = self.httpStatusCode(from: err)
                         response.status(status)
                         if let bodyData = err.bodyData {
+                            response.headers.setType("json")
                             response.send(data: try bodyData(.json))
                         }
                     } else {
@@ -505,6 +510,7 @@ extension Router {
                         let status = self.httpStatusCode(from: err)
                         response.status(status)
                         if let bodyData = err.bodyData {
+                            response.headers.setType("json")
                             response.send(data: try bodyData(.json))
                         }
                     } else {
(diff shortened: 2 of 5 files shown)


tunniclm:
Working through rebasing the code on master, there are some conflicts to resolve.

tunniclm:
Conflicts now resolved but this code requires the KituraContracts changes before it will compile.

tunniclm:
@djones6 I think I have addressed everything in the review on this PR, can you re-review for me? Thanks

djones6:
LGTM, thanks for adding those extra tests.

Package.swift
```diff
@@ -31,7 +31,7 @@ let package = Package(
     dependencies: [
         .package(url: "https://github.com/IBM-Swift/Kitura-net.git", .upToNextMinor(from: "2.0.0")),
         .package(url: "https://github.com/IBM-Swift/Kitura-TemplateEngine.git", .upToNextMinor(from: "1.7.0")),
-        .package(url: "https://github.com/IBM-Swift/KituraContracts.git", .upToNextMinor(from: "0.0.17"))
+        .package(url: "https://github.com/IBM-Swift/KituraContracts.git", .branch("add_requesterror_body"))
```
djones6: Remember to revert before this is merged :)

Sources/Kitura/CodableRouter.swift
```diff
@@ -582,6 +614,14 @@ extension Router {
                 if let err = error {
                     let status = self.httpStatusCode(from: err)
                     response.status(status)
+                    do {
+                        if let bodyData = try err.encodeBody(.json) {
+                                response.headers.setType("json")
```
ianpartridge: Indentation is off here.

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -113,337 +136,284 @@ class TestCodableRouter: KituraTest {
...
 
-    func testBasicGet() {
+    func testBasicGetSingleton() {
         router.get("/status") { (respondWith: (Status?, RequestError?) -> Void) in
             print("GET on /status")
-
             respondWith(Status("GOOD"), nil)
         }
+        router.get("/error/status") { (respondWith: (Status?, RequestError?) -> Void) in
+            print("GET on /status")
```
ianpartridge: Should be `print("GET on /error/status")`

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -113,337 +136,284 @@ class TestCodableRouter: KituraTest {
...
+            respondWith(nil, .serviceUnavailable)
+        }
+        router.get("/bodyerror/status") { (respondWith: (Status?, RequestError?) -> Void) in
+            print("GET on /status")
```
ianpartridge: `print("GET on /bodyerror/status")`

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -113,337 +135,281 @@ class TestCodableRouter: KituraTest {
...
-                XCTAssertEqual(user.id, expectedUser.id)
-
-                expectation.fulfill()
-            }, requestModifier: { request in
-                request.headers["Content-Type"] = "application/json; charset=utf-8"
-                request.write(from: userData)
-            })
-        }
+            .request("post", path: "/error/users", data: user)
+            .hasStatus(.conflict)
```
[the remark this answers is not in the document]
ianpartridge: This looks correct to me, the `"/error/users"` route responds with `nil`.

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -506,164 +465,109 @@ class TestCodableRouter: KituraTest {
...
+        let status = Status("Should not be seen")
+        router.get("/status/:id") { (id: Int, respondWith: (Status?, RequestError?) -> Void) in
+            print("GET on /status/:id that should not happen")
+            respondWith(status, nil)
+        }
+
+        buildServerTest(router, timeout: 30)
+            .request("get", path: "/status/1")
+            .hasStatus(.notFound)
+            .hasData("Cannot GET /status/1.")
```
ianpartridge: Not sure if testing the precise error message is wise - 404 seems enough.

Tests/KituraTests/TestCodableRouter.swift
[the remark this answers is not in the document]
tunniclm: It wasn't obvious from the diff, but Dave was pointing out that `.hasNoData()` wasn't being called here. Then I added it in, but his original comment remains and now it reads like he's questioning if the `.hasNoData()` should be there!

I think we all agree it should be. :)

PR #1215, opened 2018-01-31
Convert raw routing query parameters to QueryParam object

sandmman:
## Description
Now that we have added logic for query parameters into our codable routing, we can modernize the traditional routing api by providing a method to retrieve the query parameters as QueryParam type from the RouterRequest

This implementation returns an optional in case of conversion failure, but we could throw seeing as the traditional routing handling throws and will bubble up the error.

## How Has This Been Tested?
Updated the test cases. Ran locally and through travis.


ianpartridge:
Hi @sandmman is this ready to go? It looks good to me - returning an Optional seems fine.

Tests/KituraTests/TestRequests.swift
```diff
@@ -69,9 +70,17 @@ class TestRequests: KituraTest {
...
+        struct Params: QueryParams {
+            let key1: String
+            let key2: String
+            let key3: String
+        }
+
         // Test query params
         router.get("/xyz") { request, _, next in
             let expectedQueryParams = ["key1" : "value1", "key2" : "value2", "key3" : "value3"]
+            let parsedParams = request.getQueryParameters(as: Params.self)!
```
ianpartridge: I think this might be better as:

```swift
guard let parsedParams = request.getQueryParameters(as: Params.self) else {
    XCTFail("Failed to get query parameters")
    return next()
}
```

Tests/KituraTests/TestRequests.swift
```diff
@@ -83,9 +92,17 @@ class TestRequests: KituraTest {
...
+        struct Params2: QueryParams {
+            let key1: String
+            let key2: String
+            let key3: [String]
+        }
+
         // Test query parameters with multiple values assigned to a single key (array)
         router.get("/abc") { request, _, next in
             let expectedQueryParams = ["key1" : "value1", "key2" : "value2", "key3" : "value3.1,value3.2,value3.3"]
+            let parsedParams = request.getQueryParameters(as: Params2.self)!
```
ianpartridge: Same here.

Tests/KituraTests/TestRequests.swift
```diff
@@ -69,9 +70,17 @@ class TestRequests: KituraTest {
...
+        }
+
         // Test query params
         router.get("/xyz") { request, _, next in
             let expectedQueryParams = ["key1" : "value1", "key2" : "value2", "key3" : "value3"]
+            let parsedParams = request.getQueryParameters(as: Params.self)!
+            XCTAssert(parsedParams.key1 == "value1" && parsedParams.key2 == "value2" && parsedParams.key3 == "value3" )
             XCTAssertEqual(expectedQueryParams.count, request.queryParameters.count, "Unexpected number of query parameters!")
             for (key, value) in expectedQueryParams {
                 guard let v = request.queryParameters[key] else {
```
ianpartridge: Quick note that on line 87 after the XCTFail() we need to call next().

ddunn2:
@ianpartridge  resolved the merge conflicts, and addressed comments you made. 
This is now ready for another review.

sandmman:
Sorry I completely missed this message! Thanks for cleaning it up!

djones6:
@ianpartridge since the auto-jazzy is not quite behaving as we'd like right now, I suggest @sandmman just delete the Docs update from the PR and we push a separate PR for doc update (probably makes sense once the changes have been tagged).

ianpartridge:
Yep.

PR #1217, opened 2018-02-06
Bridge KituraNet.HTTPStatusCode

tunniclm:
## Description
Add a `typealias` of `HTTPStatusCode` into Kitura so that a user can use the type without also having to import KituraNet.

## Motivation and Context
While developing with Kitura, @CosmicYogi ran into unresolved symbol errors with `HTTPStatusCode`. Thus led to a discussion on Slack and this PR: https://github.com/IBM-Swift/Kitura/pull/1208.

The issue is that you need to `import KituraNet` if your code uses `HTTPStatusCode` directly.

The solution proposed by @CosmicYogi was to document the requirement in the related Kitura API documentation. This is a reasonable approach given most people will probably look at the API doc for a function if they see a compile error.

What I think would be even better, is to prevent this issue entirely so that `HTTPStatusCode` can be used even if only Kitura were imported.

## How Has This Been Tested?
Tested on macOS (in SPM and Xcode builds) and Linux, using a simple project that uses `HTTPStatusCode` and has separate test files that:
* imports Kitura only
* imports both Kitura and KituraNet

## Checklist:


tunniclm:
I tried generating Jazzy for this, but I didn't get any entry for the `typealias` at all. Not sure why.

tunniclm:
Fixed the doc generation problem (needed to add a private dummy class to keep Jazzy happy). I have added links to KituraNet.HTTPStatusCode as well.

tunniclm:
Just waiting until we have the automatic doc generation code wired up before merging this.

tunniclm:
Just waiting until we have the automatic doc generation code wired up before merging this (https://github.com/IBM-Swift/Package-Builder/pull/128)

djones6:
CI for https://github.com/IBM-Swift/Kitura/pull/1217/commits/ade5187662d0555ad461f30c2e3862062a8ba003 will fail until https://github.com/IBM-Swift/Package-Builder/pull/133 is merged.

djones6:
Github does not seem to receive the Travis CI status for a PR when the most recent commit is labelled `[ci skip]`: https://github.com/travis-ci/travis-ci/issues/2159
@ianpartridge I think we should still merge the PR as Travis for the most recent non-documentation commit (https://github.com/IBM-Swift/Kitura/pull/1217/commits/ade5187662d0555ad461f30c2e3862062a8ba003) has passed.  What do you think?

PR #1220, opened 2018-02-08
calculateExtension now appends path extension using URL.appendingPath…

siilime:
…Extension.

## Description
Completed Todo task in calculateExtension by moving to using the URL.appendingPathExtension method, (as opposed to the recommended NSString method).

## Motivation and Context
It solves the Todo listed in the original source, which was waiting on https://bugs.swift.org/browse/SR-999 to be resolved, (it has).

## How Has This Been Tested?
Ran through the tests via `swift test` and tested it on a test Kitura application locally. All 147 tests passed on macOS.

## Checklist:
Neither of the following is necessary as the existing tests and documentation should this.

### DIFF
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -183,20 +183,16 @@ public class Router {
         let fileExtension: String
         let resourceWithExtension: String
 
-        guard let resourceExtension = URL(string: template)?.pathExtension else {
+        guard let url = URL(string: template) else {
             return (fileExtension: nil, resourceWithExtension: template)
         }
 
-        if resourceExtension.isEmpty {
+        if url.pathExtension.isEmpty {
             fileExtension = defaultEngineFileExtension ?? ""
-  
-            guard let urlWithExtension = URL(string: template)?.appendingPathExtension(fileExtension) else {
-                return (fileExtension: fileExtension, resourceWithExtension: template)
-            }
             
-            resourceWithExtension = urlWithExtension.absoluteString
+            resourceWithExtension = url.appendingPathExtension(fileExtension).absoluteString
         } else {
-            fileExtension = resourceExtension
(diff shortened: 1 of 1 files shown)


Sources/Kitura/Router.swift
```diff
@@ -189,10 +189,12 @@ public class Router {
 
         if resourceExtension.isEmpty {
             fileExtension = defaultEngineFileExtension ?? ""
-            // swiftlint:disable todo
-            //TODO: Use stringByAppendingPathExtension once issue https://bugs.swift.org/browse/SR-999 is resolved
-            // swiftlint:enable todo
-            resourceWithExtension = template + "." + fileExtension
+  
+            guard let urlWithExtension = URL(string: template)?.appendingPathExtension(fileExtension) else {
+                return (fileExtension: fileExtension, resourceWithExtension: template)
```
tunniclm: There is an earlier line in the function that constructs `URL(string: template)` and we should probably only do that once. Also if it fails the first time the function will early out which means this `else` clause won't ever fire. Even so, I'm not sure the return value in the `else` clause is in keeping with the function since `fileExtension` might have a value, but it won't have been added to the end of `resourceWithExtension`.

I think it would be good to store the resource URL at line 186 instead of the extension, then that can be used in this later logic which would remove the need for the second guard (since `appendingPathExtension(_:)` returns a non-optional result.

siilime: Makes sense. I'll make the change.

siilime: @tunniclm Let me know if 7a3c91e is what you were thinking of.

ianpartridge:
Looks good to me.

tunniclm:
Thanks very much for spotting this and working through the change @siilime.
Merging.

PR #1221, opened 2018-02-21
Custom JSONEncoder for Router

tib:
## Description

* This pull request implements the following proposal: 
[KIT-0004: Custom JSONEncoder for Codable Routing](https://github.com/IBM-Swift/evolution/pull/10)

## How Has This Been Tested?

Created a new test case to see if a custom date strategy works. 
Ran `swift test` and all the test cases have been passed.

### DIFF
--- a/Tests/KituraTests/TestCodableRouter.swift
+++ b/Tests/KituraTests/TestCodableRouter.swift
@@ -40,7 +40,8 @@ class TestCodableRouter: KituraTest {
             ("testCodablePatchBodyParsing", testCodablePatchBodyParsing),
             ("testCodablePostBodyParsing", testCodablePostBodyParsing),
             ("testCodableGetQueryParameters", testCodableGetQueryParameters),
-            ("testCodableDeleteQueryParameters", testCodableDeleteQueryParameters)
+            ("testCodableDeleteQueryParameters", testCodableDeleteQueryParameters),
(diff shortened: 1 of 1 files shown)


ianpartridge:
Thanks for the PR! Can you split it into two separate PRs?  The JSONEncoder changes are logical separate from the access control changes, I think?

tib:
@ianpartridge sure, I just reverted the access level changes, this one is now for the `customJSONEncoder` block property.

EnriqueL8:
By changing a encoding strategy, we have to make sure we also change the decoding strategy. Meaning that the `JSONEncoder` and `JSONDecoder` should match. 
For the test case example of a custom date strategy it had been done [here](https://github.com/IBM-Swift/Kitura/pull/1221/files#diff-5f9c77f4430b8ae10760ad4b55dd39d0R800)
But if we think of [Kitura-Kit](https://github.com/IBM-Swift/KituraKit), we have to make sure the `JSONDecoder` for example [here](https://github.com/IBM-Swift/KituraKit/blob/a53e0ad2104678e0c80711c91ce0ce48d54b2995/Sources/KituraKit/Client.swift#L3610) matches our `JSONEncoder` in Kitura

tib:
@EnriqueL8 you are right, I just made a new [PR](https://github.com/IBM-Swift/KituraKit/pull/23) for KituraKit as well.

PR #1222, opened 2018-02-23
Refactor CodableRouter to ease code reuse

tunniclm:
## Description
Remove code duplication in the CodableRouter and expose CodableHelpers as building blocks for extensions to the CodableRouter.

## Motivation and Context
While working on a prototype codable authentication extension to the router I needed to write variations of the existing route registration functions from CodableRouter. These functions have a lot of code duplication to begin with, and with this new extension each function would need to be duplicated again to add the variations.

This motivated a refactor of these functions so they could be constructed from building block functions. These could then be exposed for use by router extensions.

## How Has This Been Tested?
Automated testing

## Checklist:

### DIFF
--- a/Sources/Kitura/CodableRouter.swift
+++ b/Sources/Kitura/CodableRouter.swift
@@ -413,6 +413,12 @@ extension Router {
 // Building blocks for Codable routing
 //
 public struct CodableHelpers {
+    /**
+     * Check if the given request has content type JSON
+     *
+     * - Parameter request: The RouterRequest to check
+     * - Returns: True if the content type of the request is application/json, false otherwise
+     */
     public static func isContentTypeJSON(_ request: RouterRequest) -> Bool {
         guard let contentType = request.headers["Content-Type"] else {
             return false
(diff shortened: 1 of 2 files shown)


tunniclm:
@djones6 Good point! Done.

Sources/Kitura/CodableRouter.swift
```diff
@@ -704,22 +402,258 @@ extension Router {
...
+     *                         closure completes execution.
+     * - Returns: The closure to pass to the codable route handler. The closure takes one argument
+     *            `(RequestError?)`.
+     *            If the argument is `nil` then the response will be considered successful, otherwise
+     *            it will be considered failed.
+     *
+     *            If successful, the HTTP status code will be set to `HTTPStatusCode.noContent` and no
+     *            body will be sent.
+     *
+     *            If failed, the HTTP status code used for the response wll be set to either the
```
djones6: typo: wll

Sources/Kitura/CodableRouter.swift
```diff
@@ -704,22 +402,258 @@ extension Router {
...
-        return (contentType.hasPrefix("application/json"))
+        return contentType.hasPrefix("application/json")
+    }
+
+    /**
+     * Get the HTTPStatusCode corresponding to the provided RequestError
+     *
+     * - Parameter from: The RequestError to map to a HTTPStatusCode
+     * - Returns: A HTTPStatusCode corresponding to the RequestError http code
+     *            if valid, or HTTPStatusCode.unknown otherwise
```
djones6: Nitpick: Should the code elements here be back-ticked?

tunniclm: Yes! I just missed them. :(

Sources/Kitura/CodableRouter.swift
tunniclm: :cry:

djones6:
Thank you for adding this excellent documentation.  Let's get this merged and the (tiny) nitpicks can be addressed later.

PR #1228, opened 2018-03-01
feat: add route for Url encoded forms

Andrew-Lees11:
## Description
update the codable router's POST handlers to detect application/x-www-form-urlencoded using the query parameters decoding logic. This allows users to send a html form to a Kitura router which will decode to a codable struct.

## How Has This Been Tested?
A new test has been added for this behaviour and this along with all other Kitura tests pass.

This was also tested manually by setting up a html form, a Kitura route and a codable struct. The form successfully sent a post request to the router which was mapped to the codable struct.

Since this is changing an existing route it will probably be worth testing the performance impact.
## Checklist:


Sources/Kitura/CodableRouter.swift
```diff
@@ -614,7 +627,7 @@ public struct CodableHelpers {
      * - Returns: An instance of `InputType` representing the decoded body data.
      */
     public static func readCodableOrSetResponseStatus<InputType: Codable>(_ inputCodableType: InputType.Type, from request: RouterRequest, response: RouterResponse) -> InputType? {
-        guard CodableHelpers.isContentTypeJSON(request) else {
+        guard CodableHelpers.isContentTypeJSON(request) || CodableHelpers.isContentTypeURLEncoded(request) else {
```
djones6: This seems okay from a performance perspective: the guard should short-circuit evaluate in the existing case (where the content is JSON) and we will only evaluate the second clause in the case where the data is not JSON.  What would this look like if we add additional content types though?

Sources/Kitura/CodableRouter.swift
```diff
@@ -624,13 +637,41 @@ public struct CodableHelpers {
...
+     */
+    public static func parseURLEncoded<InputType: Codable> (_ inputCodableType: InputType.Type, from request: RouterRequest, response: RouterResponse) throws -> InputType? {
+        var bodyData = Data()
+        let _ = try request.read(into: &bodyData)
+        guard let bodyAsString = String(data: bodyData, encoding: .utf8) else {
+            Log.error("Failed to convert the body data to String")
+            response.status(.internalServerError)
+            return nil
+        }
+        let urlKeyValuePairs = bodyAsString.urlDecodedFieldValuePairs
```
djones6: What magic is this?
Oh, I see, it's a String extension that we already defined. :+1:

Sources/Kitura/CodableRouter.swift
```diff
@@ -624,13 +637,41 @@ public struct CodableHelpers {
             return nil
         }
         do {
-            return try request.read(as: InputType.self)
+            return CodableHelpers.isContentTypeURLEncoded(request) ? try CodableHelpers.parseURLEncoded(InputType.self, from: request, response: response) : try request.read(as: InputType.self)
```
djones6: We're evaluating Content-Type a second time, and each time we call `isContentTypeX()` we have to perform a Dictionary lookup and a String compare.  There will be a performance impact for existing JSON encoding here, because we first have to ask `isContentTypeURLEncoded`, before executing the existing `request.read`.
How about we evaluate the Content-Type once, and store the result alongside the request (eg. in an enum)? Then we could write the above something like:
```swift
switch request.contentType {
    case urlEncoded: return try CodableHelpers.parseURLEncoded(InputType.self, from: request, response: response)
    case jsonEncoded: return try request.read(as: InputType.self)
}
```
Checking the value of an enum on each request would be trivial extra overhead.

In addition, the guard above could be simplified to
```swift
guard request.contentType != .unsupported else {
```

Sources/Kitura/RouterRequest.swift
```diff
@@ -184,6 +184,21 @@ public class RouterRequest {
...
+    
+    /// Read the URLEncoded body of the request as a Codable object.
+    ///
+    /// - Parameter type: Codable object to which the body of the request will be converted.
+    /// - Throws: Socket.Error if an error occurred while reading from a socket.
+    /// - Throws: `DecodingError.dataCorrupted` if values requested from the payload are corrupted, or if the given data is not valid JSON.
+    /// - Throws: An error if any value throws an error during decoding.
+    /// - Returns: The instantiated Codable object
+    public func readURLForm<T: Decodable>(as type: T.Type) throws -> T {
+        let bodyAsString = try self.readString()
```
ianpartridge: Maybe it reads better as:

```swift
guard let body = try self.readString()
      let urlKeyValuePairs = body.urlDecodedFieldValuePairs else {
    throw ()
}
return try QueryDecoder(dictionary: urlKeyValuePairs).decode(T.self)
```

Sources/Kitura/RouterRequest.swift
```diff
@@ -184,6 +184,21 @@ public class RouterRequest {
...
     }
+    
+    /// Read the URLEncoded body of the request as a Codable object.
+    ///
+    /// - Parameter type: Codable object to which the body of the request will be converted.
+    /// - Throws: Socket.Error if an error occurred while reading from a socket.
+    /// - Throws: `DecodingError.dataCorrupted` if values requested from the payload are corrupted, or if the given data is not valid JSON.
+    /// - Throws: An error if any value throws an error during decoding.
+    /// - Returns: The instantiated Codable object
+    public func readURLForm<T: Decodable>(as type: T.Type) throws -> T {
```
ianpartridge: A better name might be `readURLEncoded()`.

Andrew-Lees11:
@ianpartridge Changes are up and this can be merged once travis passes and you approve.

Sources/Kitura/CodableRouter.swift
```diff
@@ -425,7 +425,20 @@ public struct CodableHelpers {
...
+     * Check if the given request has content type x-www-form-urlencoded
+     *
+     * - Parameter request: The RouterRequest to check
+     * - Returns: True if the content type of the request is application/x-www-form-urlencoded, false otherwise
+     */
+    public static func isContentTypeURLEncoded(_ request: RouterRequest) -> Bool {
+        guard let contentType = request.headers["Content-Type"] else {
+            return false
+        }
+        return (contentType.hasPrefix("application/x-www-form-urlencoded"))
```
ianpartridge: Remove the outer parentheses.

Tests/KituraTests/KituraTestBuilder.swift
```diff
@@ -42,6 +43,7 @@ import Dispatch
 protocol RequestTestBuilder {
     func request(_ method: String, path: String) -> AssertionTestBuilder
     func request<T: Encodable>(_ method: String, path: String, data: T) -> AssertionTestBuilder
+    func request(_ method: String, path: String, URLEncodedObject: String) -> AssertionTestBuilder
```
ianpartridge: Call this `URLEncodedString`?

Tests/KituraTests/KituraTestBuilder.swift
```diff
@@ -110,6 +122,11 @@ class ServerTestBuilder: RequestTestBuilder, AssertionTestBuilder {
         requests.append(Request(test, method, path, data))
         return self
     }
+    
+    public func request(_ method: String, path: String, URLEncodedObject: String) -> AssertionTestBuilder {
```
ianpartridge: `url...` not `URL...`

Sources/Kitura/RouterRequest.swift
```diff
@@ -184,6 +184,21 @@ public class RouterRequest {
...
+    
+    /// Read the URLEncoded body of the request as a Codable object.
+    ///
+    /// - Parameter type: Codable object to which the body of the request will be converted.
+    /// - Throws: Socket.Error if an error occurred while reading from a socket.
+    /// - Throws: `DecodingError.dataCorrupted` if values requested from the payload are corrupted, or if the given data is not valid JSON.
+    /// - Throws: An error if any value throws an error during decoding.
+    /// - Returns: The instantiated Codable object
+    public func readURLEncoded<T: Decodable>(as type: T.Type) throws -> T {
+        let bodyAsString = try self.readString()
```
ianpartridge: I think this could just be `body` not `bodyAsString`. The use of `readString()` hints enough that it's a String.

Sources/Kitura/RouterRequest.swift
```diff
@@ -19,7 +19,7 @@ import Socket
 import LoggerAPI
 
 import Foundation
-
+import KituraContracts
```
ianpartridge: Add a blank line below this.

Sources/Kitura/RouterRequest.swift
```diff
@@ -184,6 +184,21 @@ public class RouterRequest {
...
+    ///
+    /// - Parameter type: Codable object to which the body of the request will be converted.
+    /// - Throws: Socket.Error if an error occurred while reading from a socket.
+    /// - Throws: `DecodingError.dataCorrupted` if values requested from the payload are corrupted, or if the given data is not valid JSON.
+    /// - Throws: An error if any value throws an error during decoding.
+    /// - Returns: The instantiated Codable object
+    public func readURLEncoded<T: Decodable>(as type: T.Type) throws -> T {
+        let bodyAsString = try self.readString()
+        guard let urlKeyValuePairs = bodyAsString?.urlDecodedFieldValuePairs else {
+            throw(Error.failedToParseRequestBody(body: bodyAsString ?? "Failed to read body as String"))
```
ianpartridge: Remove outer parentheses.

Tests/KituraTests/KituraTestBuilder.swift
```diff
@@ -84,6 +86,16 @@ class ServerTestBuilder: RequestTestBuilder, AssertionTestBuilder {
                 })
             }
         }
+        
+        init(_ test: KituraTest, _ method: String, _ path: String, URLEncodedObject: String) {
+            self.test = test
+            self.invoker = { callback in
+                test.performRequest(method, path: path, callback: callback, requestModifier: { request in
+                    request.headers["Content-Type"] = "application/x-www-form-urlencoded; charset=utf-8"
```
ianpartridge: According to https://stackoverflow.com/questions/16819502/application-x-www-form-urlencoded-and-charset-utf-8 the `application/x-www-form-urlencoded` MIME type doesn't have a `charset` parameter, so this should just be `request.headers["Content-Type"] = "application/x-www-form-urlencoded"` I think?

Andrew-Lees11:
Have done the test review comments. Have also added tests for failure case and when you send more data than the codable expected.

Sources/Kitura/RouterRequest.swift
```diff
@@ -184,6 +185,21 @@ public class RouterRequest {
         _ = try serverRequest.read(into: &data)
         return try JSONDecoder().decode(type, from: data)
     }
+    
+    /// Read the URLEncoded body of the request as a Codable object.
+    ///
+    /// - Parameter type: Codable object to which the body of the request will be converted.
```
ianpartridge: This parameter is called `as` not `type` (the doc is wrong on the function above too...)

Tests/KituraTests/KituraTestBuilder.swift
```diff
@@ -84,6 +86,16 @@ class ServerTestBuilder: RequestTestBuilder, AssertionTestBuilder {
                 })
             }
         }
+        
+        init(_ test: KituraTest, _ method: String, _ path: String, _ URLEncodedObject: String) {
```
ianpartridge: URLEncodedObject - > urlEncodedString I think, to be consistent with the `request()` function.

Tests/KituraTests/KituraTestBuilder.swift
```diff
@@ -84,6 +86,16 @@ class ServerTestBuilder: RequestTestBuilder, AssertionTestBuilder {
                 })
             }
         }
+        
+        init(_ test: KituraTest, _ method: String, _ path: String, _ URLEncodedString: String) {
```
ianpartridge: Sorry, can this be urlEncodedString not URLEncodedString, otherwise it looks like a reference to a type.

djones6:
FWIW, there was no measurable performance penalty for a simple benchmark using Codable routing:
```
               | Throughput (req/s)      | CPU (%) | Mem (kb)     | Latency (ms)                   | good
Implementation | Average    | Max        | Average | Avg peak RSS | Average  | 99%      | Max      | iters
---------------|------------|------------|---------|--------------|----------|----------|----------|-------
          Base |    34122.9 |    34337.7 |    98.4 |        29125 |      3.7 |      4.2 |     20.1 |     5
        pr1228 |    34158.5 |    34385.6 |    98.2 |        28717 |      3.7 |      4.2 |     21.3 |     5
```

Sources/Kitura/CodableRouter.swift
```diff
@@ -425,7 +425,20 @@ public struct CodableHelpers {
...
+     * Check if the given request has content type x-www-form-urlencoded
+     *
+     * - Parameter request: The RouterRequest to check
+     * - Returns: True if the content type of the request is application/x-www-form-urlencoded, false otherwise
+     */
+    public static func isContentTypeURLEncoded(_ request: RouterRequest) -> Bool {
+        guard let contentType = request.headers["Content-Type"] else {
+            return false
+        }
+        return contentType.hasPrefix("application/x-www-form-urlencoded")
```
djones6: This is executed twice on the critical path (by `read()`, and by `readCodableOrSetResponseStatus()` which itself calls `read()`).  Each execution involves a dictionary lookup and a call to `String.hasPrefix()`, both of which involve String comparison.

It would be better to cache this state against the Request itself (eg. in an enum), particularly if we end up supporting more content types in the future.  For now though, it's a trivial extra cost and we can make a note to fix it in the future.

Sources/Kitura/RouterRequest.swift
```diff
@@ -172,19 +173,27 @@ public class RouterRequest {
...
+    /// Read the body of the request as a Codable object. It can decode JSON or URLEncoded forms.
+    /// It chooses the decoder by looking up the Content-Type header.
+    /// If there is no header it defaults to JSONDecoder.
+    /// - Parameter as: Codable object to which the body of the request will be converted.
     /// - Throws: Socket.Error if an error occurred while reading from a socket.
     /// - Throws: `DecodingError.dataCorrupted` if values requested from the payload are corrupted, or if the given data is not valid JSON.
     /// - Throws: An error if any value throws an error during decoding.
     /// - Returns: The instantiated Codable object
     public func read<T: Decodable>(as type: T.Type) throws -> T {
+        if CodableHelpers.isContentTypeURLEncoded(self) {
```
djones6: If the content type was stored on Request, we could simply switch on it.  (see my comment above)

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -162,6 +165,26 @@ class TestCodableRouter: KituraTest {
             .hasStatus(.conflict)
             .hasContentType(withPrefix: "application/json")
             .hasData(Conflict(on: "id"))
+            
+            .request("post", path: "/urlencoded", urlEncodedString: "id=4&name=David")
+            .hasStatus(.created)
+            .hasContentType(withPrefix: "application/json")
+            .hasData(user)
```
djones6: :+1: for nicely concise and expressive tests.

svanimpe:
Is there a way to use this in server-side web apps (routes that receive a form and render a HTML response)?

ianpartridge:
I'm afraid the Codable routing handlers are still JSON and REST-centric...

Andrew-Lees11:
@svanimpe Yes. If you make a cobable object which maps to your HTML form and have it post to your route you can use this and then use the object to render the response. I will be writting a blog post describing this process and update our [FoodTracker Stencil](https://github.com/Andrew-Lees11/FoodTrackerBackend/blob/master/StencilFrontend.md) tutorial this week.

Issue #1216, opened 2018-01-31
Failure to call callbacks leading to memory leaks.


drewmccormack:
We have a Kitura server deployed, and have to restart it every 6 hours to stop it using excessive memory. It steadily climbs over time. Clearly a leak.

We thought it could be our code, so we have investigated with Instruments on the command line, using the Leaks instrument.

We were able to track down and fix some leaks, but they were not in our code, but in Kitura itself. In particular, there are a number of places where callbacks are not called, leading walkers to be incomplete and left hanging in memory due the resulting retain cycle.

Here is one example:
https://github.com/IBM-Swift/Kitura/blob/3d42897fa83537b97592512a7a85af61e9ce53ea/Sources/Kitura/Router.swift#L365-L383

The `process` function of `Router`. There are two paths through this function that do not trigger the callback: 

1. The guard at the top
2. The first branch of the `if`

Why is that so important? Because the `process` function is called by the `handle` function in the same class, and an assumption is made there that the callback will always trigger, leading to a call to `popRouter`. If that call is not made, there is a leak.

This may be the most serious problem we found, but was not the only one. In many other places, there are paths through functions that fail to call the `next` callback. Not doing this can again leave walkers and other asynchronous chains hanging, and leading to a leak.

Here are some of the places we found missing `next` calls:

- `Router` class, `handle` func. Both `guard` fail to call `next`
- `RouterElement`, `process` func. `guard` at top
- `RouterElement`, `performSimpleMatch` func. `guard` at top
- `BodyParser`, `handle` func. Both `guard`s
- In KituraCredentials, there were several paths in `Credentials` class, `handle` and `authenticate` func that do not trigger `next`.

Making these changes, we were able to reduce the leaks considerably, although we weren't able to track leaks that arose during 401 authorization errors. So there are likely more places in Kitura not abiding by the callback rules. An audit of functions using callbacks may be useful, and setting up some sort of leak testing.

seabaylea:
Hi @drewmccormack. Thanks for the thorough write up! @djones6 can you investigate?

nethraravindran:
@drewmccormack Thanks for the write up. With your suggestions, I have created a local branch.

https://github.com/nethraravindran/Kitura/tree/router-memoryleak

With this, I could observe the same minimal amount of leak. I request you to test once and check if the leak still persists. Also it would be great if you could give us the scenario to produce a considerable amount of leak. Thanks!

drewmccormack:
[...]

PR #1232, opened 2018-03-09
Fix for failure to call callbacks leading to memory leaks

nethraravindran:
## Description

Code paths in ```Router``` and ```RouterElement``` failed to call callbacks which leaves walkers and other asynchronous chains to be hanging and leak memory.  

## Motivation and Context

Issue: https://github.com/IBM-Swift/Kitura/issues/1216

## Checklist:


Sources/Kitura/RouterElement.swift
```diff
@@ -146,7 +147,10 @@ class RouterElement {
     /// - Parameter next: The closure called to invoke the next handler or middleware
     ///                     associated with the request.
     private func performSimpleMatch(path: String, request: RouterRequest, response: RouterResponse, next: @escaping () -> Void) {
-        guard let pattern = pattern else { return }
+        guard let pattern = pattern else {
+            next()
+	    return
```
ianpartridge: There's a tab hidden in this line I'm afraid.

djones6:
@ianpartridge It looks like `sendIfFound()` calls `response.end()`, so we should not then call `next()`.

PR #1237, opened 2018-03-20
Store URL encoded key-value pairs as [String: [String]]

NocturnalSolutions:
… rather than a [String: String].

## Description

When a GET request is done with query parameters, such as `http://example.com/search?query=foobar`, those parameters are accessible in the `.queryParameters` property of the RouterRequest object in the handler. If a POST is done with the default "application/x-www-form-urlencoded" enctype, those parameters are accessible by calling `.body?.asURLEncoded` on the RouterRequest object. Both of these are [String: String] dictionaries. These changes change them to be [String: [String]] dictionaries; in other words, rather than each key having a single string value, they have an array of strings.

A [PR for KituraContracts](https://github.com/IBM-Swift/KituraContracts/pull/16) will also need to be accepted to reflect these changes. For now, if you clone my branch to test, you'll need to change the line for KituraContracts in Kitura's `Package.swift` to:

```swift
        .package(url: "https://github.com/NocturnalSolutions/KituraContracts.git", .branch("UrlEncodingValuesArray")),
```

## Motivation and Context

 The code that decodes the URL encoding to create these dictionaries is in [String+Extensions.swift](https://github.com/IBM-Swift/Kitura/blob/master/Sources/Kitura/String%2BExtensions.swift).

Consider a form such as: 

```html
<h1>What are your favorite pizza toppings?</h1>
<form method="get">
  <label><input type="checkbox" name="toppings" value="extraCheese" />Extra cheese</label>
  <label><input type="checkbox" name="toppings" value="pepperoni" />Pepperoni</label>
  <label><input type="checkbox" name="toppings" value="pineapple" />Pineapple</label>
  <label><input type="checkbox" name="toppings" value="anchovies" />Anchovies</label>
  <input type="submit" />
</form>
```

If you were to select both "Extra cheese" and "Pineapple," the resulting URL-encoded string would look something like `toppings=extraCheese&toppings=pineapple`. So there are two values for the same key.

However, Kitura will only allow a single string value for each key. So Kitura works around this by using a comma to concatenate all of the values. The result is that `request.queryParameters["toppings"]` will be `"extraCheese,pineapple"`.

I don't like this. For one, it makes it impossible to simply count how many values you have for a key without doing something like `.split(separator: ",").count`. For another, what if a value itself has a comma? Now we're in painful territory.

Unfortunately, this is a fairly significant change that will break existing sites that use forms. Basically, instead of using `let name = request.queryParameters["name"]`, they'll have to use `let name = request.queryParameters["name"]?.first` if they're just expecting a single value in the request. On the other hand, for those if they're doing silliness like `let names = request.queryParameters["name"]?.split(separator: ",")`, it'll just become `let names = request.queryParameters["name"]`.

I'm willing to bet Swift has a way to make a thing which acts like a dictionary if you do a subscript operation on it but acts like a string otherwise, but I think that might be overengineering in this case.

## How Has This Been Tested?

Existing tests have been adjusted to work with the change.

## Checklist:

### DIFF
--- a/Sources/Kitura/RouterRequest.swift
+++ b/Sources/Kitura/RouterRequest.swift
@@ -138,7 +138,7 @@ public class RouterRequest {
     public internal(set) var parameters: [String:String] = [:]
 
     /// List of query parameters.
-    public lazy var queryParameters: [String:String] = { [unowned self] in
+    public lazy var queryParameters: [String:[String]] = { [unowned self] in
         return self.urlURL.query?.urlDecodedFieldValuePairs ?? [:]
         }()
 
--- a/Sources/Kitura/RouterResponse.swift
+++ b/Sources/Kitura/RouterResponse.swift
@@ -507,7 +507,7 @@ extension RouterResponse {
 
         let jsonStr = String(data: try encoder.encode(jsonp), encoding: .utf8)!
 
-        let taintedJSCallbackName = request.queryParameters[callbackParameter]
+        let taintedJSCallbackName = request.queryParameters[callbackParameter]?.first
 
         if let jsCallbackName = validJsonpCallbackName(taintedJSCallbackName) {
             headers.setType("js")
--- a/Sources/Kitura/String+Extensions.swift
+++ b/Sources/Kitura/String+Extensions.swift
@@ -18,11 +18,12 @@ import Foundation
 import LoggerAPI
 
 /// String Utils
+
 extension String {
 
     /// Parses percent encoded string into query parameters
-    var urlDecodedFieldValuePairs: [String : String] {
-        var result: [String:String] = [:]
+    var urlDecodedFieldValuePairs: [String: [String]] {
+        var result: [String: [String]] = [:]
 
         for item in self.components(separatedBy: "&") {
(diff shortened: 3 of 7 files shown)


NocturnalSolutions:
The CI failing is expected as, as mentioned above, this PR is recursively dependent on https://github.com/IBM-Swift/KituraContracts/pull/16.

ianpartridge:
Thank you for the PR!  As you mentioned, this is unfortunately an API breaking change so merging it would require us to tag Kitura 3.0 - which we don't plan to do any time soon.

Is there some way we could add the functionality you want in an additive way?  i.e. by leaving `request.queryParameters` unchanged and adding new API instead?

NocturnalSolutions:
> Is there some way we could add the functionality you want in an additive way? i.e. by leaving request.queryParameters unchanged and adding new API instead?

To be clear, are you speaking of implementing it another way inside of Kitura, or as a separate package/module that wouldn't involve patches to Kitura?

ianpartridge:
I think inside of Kitura would probably be fine.

djones6:
Closing as the successor #1238 is now merged.

PR #1238, opened 2018-03-22
Allow [String: [String]] dictionaries to be resultant of parsing URL-encoded strings

NocturnalSolutions:
…rather than [String: String] dictionaries where multiple values for keys are comma-separated.

## Description

A urlDecodedFieldMultiValuePairs() method is added to String alongside the existing urlDecodedFieldValuePairs(). Additionally, a BodyParserMultiValue class is added as a subclass of the existing BodyParser class. (Both of these are kind of ugly since the existing code isn't written to be flexible in this regard…)

## Motivation and Context

I had a PR at https://github.com/IBM-Swift/Kitura/pull/1237 that would replace existing code related to decoding URL-encoding GET query strings and POST bodies so that `foo=bar&foo=baz` would result in `["foo": ["bar", "baz"]]` rather than `["foo": "bar,baz"]` as is the current behavior. As this would break pretty much all existing sites that handle such things, @ianpartridge asked me to try again in a way that wouldn't cause breakage. This is the result.

From the other PR:

> The code that decodes the URL encoding to create these dictionaries is in [String+Extensions.swift](https://github.com/IBM-Swift/Kitura/blob/master/Sources/Kitura/String%2BExtensions.swift).
> 
> Consider a form such as: 
> 
> ```html
> <h1>What are your favorite pizza toppings?</h1>
> <form method="get">
>   <label><input type="checkbox" name="toppings" value="extraCheese" />Extra cheese</label>
>   <label><input type="checkbox" name="toppings" value="pepperoni" />Pepperoni</label>
>   <label><input type="checkbox" name="toppings" value="pineapple" />Pineapple</label>
>   <label><input type="checkbox" name="toppings" value="anchovies" />Anchovies</label>
>   <input type="submit" />
> </form>
> ```
> 
> If you were to select both "Extra cheese" and "Pineapple," the resulting URL-encoded string would look something like `toppings=extraCheese&toppings=pineapple`. So there are two values for the same key.
> 
> However, Kitura will only allow a single string value for each key. So Kitura works around this by using a comma to concatenate all of the values. The result is that `request.queryParameters["toppings"]` will be `"extraCheese,pineapple"`.
> 
> I don't like this. For one, it makes it impossible to simply count how many values you have for a key without doing something like `.split(separator: ",").count`. For another, what if a value itself has a comma? Now we're in painful territory.

With this PR, `request.queryParametersMultiValues` can be used in place of `request.queryParameters` to get values for a key as a `[String]`. Similarly, for POST forms, the following can be done:

```swift
router.post("/submit", middleware: BodyParserMultiValue()) // instead of BodyParser()
router.post("/submit") { request, response, next in
    if let postValues = request.body?.asURLEncodedMultiValue { // instead of .asURLEncoded
        // postValues is a [String: [String]] now
```

## How Has This Been Tested?

Test cases have been added for both GET and POST requests.

## Checklist:


Tests/KituraTests/TestRequests.swift
```diff
@@ -97,6 +97,21 @@ class TestRequests: KituraTest {
...
+            for (key, value) in expectedQueryParams {
+                guard let v = request.queryParametersMultiValues[key] else {
+                    XCTFail("Query parameter \(key) was nil!")
+                    return
+                }
+                XCTAssertEqual(v, value)
+            }
+            next()
+        }
+
```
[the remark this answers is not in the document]
ianpartridge: Yes restructuring would be nice but we can do that separately.

Sources/Kitura/String+Extensions.swift
```diff
@@ -20,9 +20,20 @@ import LoggerAPI
 /// String Utils
 extension String {
 
-    /// Parses percent encoded string into query parameters
+    /// Parses percent encoded string into query parameters with comma-separated
+    /// values.
     var urlDecodedFieldValuePairs: [String : String] {
```
ianpartridge: Although this does now look like a slow path (it uses `urlDecodedFieldMultiValuePairs` then joins them with commas), I suspect it may be faster than before as we no longer do string interpolation which is known to be slow.

Having said that it would be nice to know what the performance impact is. @NocturnalSolutions did you measure any difference? @djones6 do you have any perf tests that would cover this?

ianpartridge: https://github.com/wg/wrk is pretty good - point it at a Kitura app with a route that exercises this codepath...

NocturnalSolutions:
Pushed tweaks to avoid force unwrapping as per @djones6 's concerns. (Thanks to ch1ck3n in #swift-lang (Freenode) for reminding me of the `default` trick.) 

Perf tested again under similar circumstances:

```
> wrk -t8 -c100 -d1m 'http://localhost:8080/?foo=bar&baz=qux&multival=123&multival=234'
Running 1m test @ http://localhost:8080/?foo=bar&baz=qux&multival=123&multival=234
  8 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     8.57ms    6.65ms 191.45ms   97.70%
    Req/Sec     1.50k   107.23     2.29k    74.62%
  715253 requests in 1.00m, 137.79MB read
Requests/sec:  11907.66
Transfer/sec:      2.29MB
```

Huh, so that's actually faster than my test with the unpatched code base, which doesn't make sense unless the speed change is actually negligible and all these results are just falling within a margin of error.

djones6:
@NocturnalSolutions There is a fair margin of error with this sort of testing - your best and worst results are 2% apart, which I'd say is inconclusive based on a single execution of each test.

A way to get around that is to execute each test multiple times (before, after, before, after...) and compare the average of the befores with the average of the afters.

The test itself that you are performing looks fine, but I'd say an equally interesting case would be where you have no multi values at all:  the new code creates an empty array, appends a single element, then invokes `.joined` on the single-element array.  (I don't know what degree of fast paths the stdlib / compiler apply in this case).

I'd also expect that the level of regression a change like this could cause to a Kitura workload is very small - this is small fry considering the total cost of reading/writing to the socket, constructing the Request, Response objects, HTTP parsing, headers container and all the other bits and bobs involved in serving a request.

If we want to measure which approach is faster, for the sake of interest, then you could extract the before/after code into a standalone testcase and time it running in a loop.

Oh, one other (crucial) detail: make sure you compile with `swift build -c release` (or `swiftc -O` for simple scripts) when comparing performance: release mode enables compiler optimizations that are not applied to the default (debug) build, which can dramatically affect a comparison.

NocturnalSolutions:
Okay, here's some results from going back-and-forth between patched and unpatched packages and testing in between, and using that `-c release` bit (which made a rather significant difference). The command run this time was `http://localhost:8080/?foo=bar&baz=qux&multival=123` - so no actual multiple value GET parameter this time. Also, unlike my previous testing, I didn't quit all my other Mac apps besides Terminal this time.

Before results:

```
Running 1m test @ http://localhost:8080/?foo=bar&baz=qux&multival=123
  8 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     7.15ms   12.03ms 230.47ms   94.67%
    Req/Sec     2.44k   319.24     4.05k    78.81%
  1164897 requests in 1.00m, 219.96MB read
Requests/sec:  19390.12
Transfer/sec:      3.66MB

Running 1m test @ http://localhost:8080/?foo=bar&baz=qux&multival=123
  8 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     7.69ms   13.80ms 255.25ms   94.61%
    Req/Sec     2.37k   371.07     4.00k    84.33%
  1134102 requests in 1.00m, 214.15MB read
Requests/sec:  18876.50
Transfer/sec:      3.56MB

  8 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     7.81ms   14.81ms 234.57ms   94.50%
    Req/Sec     2.45k   298.84     4.76k    76.06%
  1169577 requests in 1.00m, 220.85MB read
Requests/sec:  19468.90
Transfer/sec:      3.68MB
```

Average req/sec: 19,245.17

After patching:

```
Running 1m test @ http://localhost:8080/?foo=bar&baz=qux&multival=123
  8 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     7.38ms   12.23ms 235.99ms   95.33%
    Req/Sec     2.25k   316.03     6.98k    82.35%
  1074179 requests in 1.00m, 202.83MB read
Requests/sec:  17883.98
Transfer/sec:      3.38MB

Running 1m test @ http://localhost:8080/?foo=bar&baz=qux&multival=123
  8 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     6.86ms   10.57ms 224.09ms   95.64%
    Req/Sec     2.31k   310.92     3.68k    81.40%
  1102305 requests in 1.00m, 208.15MB read
Requests/sec:  18356.73
Transfer/sec:      3.47MB

Running 1m test @ http://localhost:8080/?foo=bar&baz=qux&multival=123
  8 threads and 100 connections
  Thread Stats   Avg      Stdev     Max   +/- Stdev
    Latency     7.89ms   13.85ms 225.28ms   94.38%
    Req/Sec     2.33k   304.64     4.25k    77.88%
  1113263 requests in 1.00m, 210.21MB read
Requests/sec:  18535.43
Transfer/sec:      3.50MB
```

Average req/sec: 18,258.71

Just as an aside, I know this is an artificial test with no real network use, but still, over 18 thousand requests per second on my oldish MBP? That's an unimaginably big number to me, and I don't think I could get similar numbers with a PHP-based stack on the same hardware without a lot more work. Great job, Swift and Kitura teams!

djones6:
I ran some measurements on a Linux machine locally with the same route as @NocturnalSolutions - there is regression, the amount of regression depends on the query params specified but is surprising:

- No query params:  no regression (70.8k vs. 70.9k)
- `?a=b`:  2% regression (43.9k vs. 44.8k)
- `?a=b&c=d&a=c&a=d`:  3% regression (30.3k vs. 31.3k)
- `?a=b&a=c&a=d&a=e&a=f&a=g&a=h&a=i&a=j&a=k`:  1.5% regression (21k vs. 21.3k)
- `?a=b&c=d&e=f&g=h&i=j&k=l&m=n&o=p&q=r&s=t`: 6% regression (17.6k vs. 18.7k)

The regression increases with the number of distinct query params:  a single query param with 10 values shows only 1.5% regression, whereas 10 query params each with a single value regresses by 6%.

I'm particularly surprised that this code could cause such a significant regression, but it does indicate that going via `urlDecodedFieldMultiValuePairs` is substantially more expensive for single value params than the existing code.  As a quick experiment, I tried replacing the `values.joined(separator:)` with `values.first` but this made no measurable difference to the 6% case above,  so it seems the whole cost is in building the intermediate array representation while computing `urlDecodedFieldMultiValuePairs`.

For this reason I think it would be worth the extra bit of code duplication to maintain the two as independent functions to avoid the regression for the common case (single value params).

NocturnalSolutions:
Okay, pushed a change to duplicate the code in the relevant part. (I also gave up on trying to get the encoding/decoding bit in KituraContracts to play along. I don't think I had that working 100% in the first place. At any rate, I guess it can be saved for another day. So now you don't have to also patch KituraContracts to test this stuff.) I think there's deduplication that can be done in other ways, but for now, at commit 2e119e3, with `wrk -t8 -c100 -d1m 'http://localhost:8080/?foo=bar&baz=qux&multival=123&a=b&c=d&e=f&g=h&i=j&k=l&mno=pqrst' `, I get 10,841.50 reqs/sec before the patch and 11,060.16 reqs/sec after - probably due to statistical noise.

NocturnalSolutions:
Ah, that's better. Before: 11154.13, after: 11132.90.

NocturnalSolutions:
Ah, that's better. Before: 11154.13, after: 11132.90.

ianpartridge:
@djones6 any further thoughts on this?

djones6:
I confirmed on Linux that this is now performance-neutral, so from that perspective I think it is fine to go ahead.

PR #1242, opened 2018-04-11
Relax Codable Router type requirements

### LINKED ISSUE #1235: CodableClosure should use Decodable -> Encodable
## Context and Description
At the moment the CodableClosure forces the users that the request-object and the response-object is Codable. So if the response is only Encodable, it is not possible to use the Codable-Route.

### Environment Details
MacOS 10.13.3 (17D102)

Apple Swift version 4.0.3 (swiftlang-900.0.74.1 clang-900.0.39.2)
XCode Version 9.2 (9C40b)
Kitora Version 2.2.0

### Steps to Reproduce
1) 1. create a codable-route, for example
```
router.post("/users", handler: { (request: MyRequest, respondWith: (MyResponse?, RequestError?) -> Void) in
    respondWith(nil, nil)
})
```
2) MyRequest implements Decodable, MyResponse implements Encodable
3) Doesn't compile
4) Adding Codable to MyRequest and MyResponse will work

djones6:
## Description
Implementation of Kitura Evolution proposal: [Relax Type Constraints for Codable Routing](https://github.com/IBM-Swift/evolution/pull/11)

Resolves #1235 by relaxing the type requirements on the Codable Router to Decodable for input types, and Encodable for output types.

This also requires a corresponding Pull Request in KituraContracts: IBM-Swift/KituraContracts#18

## Motivation and Context

See #1235.

The Codable Router was designed with a symmetry in mind for types to be shared between the client and server side (KituraKit <-> Kitura), which requires those types to be both Encodable and Decodable.

In addition, for types that contain purely Codable attributes, conformance to Codable can be synthesized by the compiler.  However, there are cases where conformance must be implemented manually.

For cases where the user only wants to use type safety in a single direction (such as Decodable for a type that will only be received), we currently require them to implement conformance to Encodable (at least dummy conformance to satisfy the type requirement).

### Notable changes to behaviour
In order to facilitate the receiving of a `Decodable` type in a PUT or POST, without requiring that we respond with the same type (or some dummy type) conforms to `Encodable`, it is necessary to allow Codable responses of `(nil, .someSuccessStatus)`.

In the current implementation I have relaxed this to also permit  `(nil, nil)`, where the nil `RequestError` signifies success (with the default code for that method), and the nil `Encodable` signifies that we wish to return no data.

Previously, such a response would have failed (due to trying to JSON encode a `nil`), and resulted in logging in error (`Could not encode result: invalidValue(nil, ....`) and returning `.internalServerError` status.

## Checklist:


djones6:
@ianpartridge I've rebased this (and IBM-Swift/KituraContracts#18) on the latest from master, and proposed a solution for the issue of POST / PUT which I have detailed in 'Notable changes to behaviour' in the description above.

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -135,6 +139,40 @@ class TestCodableRouter: KituraTest {
         }
     }
 
+    enum OnlyEncodable: Encodable {
+	case data(String)
```
ianpartridge: Extra space needed.

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -135,6 +139,40 @@ class TestCodableRouter: KituraTest {
...
+            var container = encoder.container(keyedBy: CodingKeys.self)
+            switch self {
+            case .data(let value):
+                try container.encode(value, forKey: .data)
+            }
+        }
+    }
+
+    enum OnlyDecodable: Decodable {
+	case data(String)
```
ianpartridge: Extra space needed.

djones6:
@ianpartridge done.

PR #1243, opened 2018-04-11
Move mergeParameters docs to correct location

ddunn2:
Currently an example of the use of mergeParameters exists within the overall Router [docs](https://ibm-swift.github.io/Kitura/Classes/Router.html). This PR moves this example into the correct location, where mergeParameters is discussed in the [docs](https://ibm-swift.github.io/Kitura/Classes/Router.html#/s:6Kitura6RouterCACSb15mergeParameters_tcfc)


Sources/Kitura/Router.swift
```diff
@@ -93,14 +72,28 @@ public class Router {
...
     fileprivate var parameterHandlers = [String : [RouterParameterHandler]]()
 
-    /// Initialize a `Router` instance. The optional `mergeParameters` parameter allows you to control whether
-    /// the router will be able to access parameters matched in its parent router.
-    /// ### Usage Example: ###
+    /// When initialising a `Router`, `mergeParameters` allows you to control whether
+    /// the router will be able to access parameters matched in its parent router. For instance, in the example below
+    /// if `mergeParameters` is set to `true`, `GET /Hello/Alex` will return "Hello Alex", but if set to `false`
+    /// the `greeting` parameter will not be accessible and it will return just " Alex".
     /// ```swift
```
alexwishart: As `mergeParameters` is optional, I still think we should keep the simple code example that we had (i.e. just `let router = Router()`)  under a "Usage Example:" header to follow the same convention as all the other functions. It's the most common way that the function is used so think we need to have it separated from the optional `mergeParameters` so people don't get caught up in the extra stuff?

Could potentially add a "Using mergeParameters" subheader under "Usage Example" and the original code and description with the extra code snippet and description? 

@ianpartridge would appreciate your opinion on this too 🙂

ianpartridge: ^^ what Alex said :)

ddunn2:
Addressed review comments

ddunn2:
Closing this PR as it's clear more than just 'moving the `mergeParameters` docs to the correct location' is needed. Will review the current API docs, and decide the best way to include the information needed and open a new PR when I have that.

PR #1250, opened 2018-04-15
Update Router Templating API docs

ddunn2:
## Description
Update the `TemplateEngine` docs to provide more information regarding edge cases such as registering multiple different template engines to use the same extension. 

## Motivation and Context
Makes the API docs clearer for users.

### DIFF
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -27,7 +27,7 @@ import KituraTemplateEngine
 ///   - Routing requests to closures of type `RouterHandler`
 ///   - Routing requests to the handle function of classes that implement the
 ///    `RouterMiddleware` protocol.
-///   - Routing requests to a template engine to generate the appropriate output.
+///   - Routing requests to a `TemplateEngine` to generate the appropriate output.
 ///   - Serving the landing page when someone makes an HTTP request with a path of slash (/).
 ///
 /// When initialising a `Router`, `mergeParameters` allows you to control whether
@@ -126,14 +126,19 @@ public class Router {
 
     // MARK: Template Engine
 
-    /// Register a templating engine. The templating engine will handle files in the `viewsPath`
-    /// that match the file extension it supports.
+    /// Register a template engine to a given router instance.
+    /// A template engine allows rendering of documents using static templates.
+    ///
+    /// By default the templating engine will handle files in the `./Views` directory
+    /// that match the file extension it supports. You can change this default location using the `viewsPath` property.
     /// ### Usage Example: ###
     /// ```swift
     /// let router = Router()
     /// router.add(templateEngine: MyTemplateEngine())
     /// router.add(templateEngine: MyOtherTemplateEngine(), forFileExtensions: ["html"], useDefaultFileExtension: false)
     /// ```
+    /// If multiple different template engines are registered for the same extension, the template engine
+    /// that is registered last will be the one that attempts to render all template files with the chosen extension.
     /// - Parameter templateEngine: The templating engine to register.
     /// - Parameter forFileExtensions: The extensions of the files to apply the template engine on.
     /// - Parameter useDefaultFileExtension: The flag to specify if the default file extension of the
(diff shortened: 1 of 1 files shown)


Sources/Kitura/Router.swift
```diff
@@ -126,14 +126,22 @@ public class Router {
...
+    /// that match the file extension it supports. You can change this default location using the `viewsPath` property.
     /// ### Usage Example: ###
     /// ```swift
     /// let router = Router()
+    /// //Register a template engine to a given router instance
     /// router.add(templateEngine: MyTemplateEngine())
+    /// //Register a template engine to a given router instance using a custom file extension
     /// router.add(templateEngine: MyOtherTemplateEngine(), forFileExtensions: ["html"], useDefaultFileExtension: false)
+    /// //Change the location a template engine will look for template files in any given router instance
+    /// router.viewsPath = "New_Custom_Location"
```
alexwishart: This doesn't need to be here, it's already been covered in https://ibm-swift.github.io/Kitura/Classes/Router.html#/s:6Kitura6RouterC9viewsPathSSv

Sources/Kitura/Router.swift
```diff
@@ -156,6 +164,9 @@ public class Router {
     /// let router = Router()
     /// router.setDefault(templateEngine: MyTemplateEngine())
     /// ```
+    /// If the template engine doesn't provide a default extension you can provide one using
+    /// `add(templateEngine:forFileExtensions:useDefaultFileExtension:)`. If a router instance doesn't
+    /// have a template engine registered that can render the given template file an error is thrown.
```
alexwishart: Could you add in what the error is, then people can recognise that it's the "template engine missing" error and not something else.

Sources/Kitura/Router.swift
```diff
@@ -126,14 +126,22 @@ public class Router {
 
     // MARK: Template Engine
 
-    /// Register a templating engine. The templating engine will handle files in the `viewsPath`
-    /// that match the file extension it supports.
+    /// A template engine allows rendering of documents using static templates.
```
alexwishart: Convention in the other functions is that the first thing you see explicitly says what the specific function does e.g. "Set up a sub router to handle requests." With any extra explanation coming after. I think if you just move the first comment line from your code (i.e. "Register a template engine to a given router instance") to the top here it's fine 🙂

ddunn2: I'll look into how to best phrase this as 'Register a templating engine' just felt out of place there. I like your idea of moving the comment up to the top, that could work with a little rewording :).

Sources/Kitura/Router.swift
```diff
@@ -126,14 +126,22 @@ public class Router {
...
+    ///
+    /// By default the templating engine will handle files in the `./Views` directory
+    /// that match the file extension it supports. You can change this default location using the `viewsPath` property.
     /// ### Usage Example: ###
     /// ```swift
     /// let router = Router()
+    /// //Register a template engine to a given router instance
```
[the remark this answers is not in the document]
ddunn2: I think having no comments is the 'old' convention. In the Codable routing API we have comments. I will however follow whichever convention, so will remove the comments.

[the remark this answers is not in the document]
alexwishart: @helenmasters can you confirm r.e. comments/no comments? Maybe we need to create some documentation somewhere on how we want the APIs to look (although I think we'd started to use [Kitura-Cache](https://github.com/IBM-Swift/Kitura-Cache ) as an example?)

Sources/Kitura/Router.swift
```diff
@@ -156,6 +161,10 @@ public class Router {
     /// let router = Router()
     /// router.setDefault(templateEngine: MyTemplateEngine())
     /// ```
+    /// If the template engine doesn't provide a default extension you can provide one using
+    /// `add(templateEngine:forFileExtensions:useDefaultFileExtension:)`. If a router instance doesn't
+    /// have a template engine registered that can render the given template file an
```
alexwishart: "an" should be "a"

Sources/Kitura/Router.swift
```diff
@@ -156,6 +161,10 @@ public class Router {
...
+    /// `No template engine defined for extension` error is thrown.
```
ianpartridge: This is a `TemplatingError` I think - only use backticks for code.

ddunn2:
Addressed review comments, will ask @helenmasters to review as well.

PR #1255, opened 2018-04-20
Make `RouterResponse.request` public so it can be used in extensions

svanimpe:
I'd like to propose this small change.

This greatly simplifies my use case:

- My app renders complex HTML pages.
- To keep routers small and focused, the creation of a rendering context (`[String: Any]`) is done using separate view model types.
- My Stencil templates use inheritance. As a result, all view models have a common base context.
- To avoid code duplication, the data for this base context is prepared by a middleware and stored in `request.userInfo`.
- To render a page, I call a custom extension method on `RouterResponse` that can render a view model. This method pulls the base context from `request.userInfo`, merges it with the contents of the view model and uses the result as a rendering context for the template:

```swift
@discardableResult
func render(_ resource: String, viewModel: ViewModel) throws -> RouterResponse {
    let baseContext = request.userInfo
    return try self.render(resource, context: baseContext.merging(viewModel.contents))
}
```

However, this only works when `request` is public. Without this change, I would have to pass the request to my extension method, which feels silly, because the instance already has the request as a property; it's just not public.

### DIFF
--- a/Sources/Kitura/RouterResponse.swift
+++ b/Sources/Kitura/RouterResponse.swift
@@ -74,7 +74,7 @@ public class RouterResponse {
     private var routerStack: Stack<Router>
 
     /// The associated request
-    let request: RouterRequest
+    public let request: RouterRequest
 
     /// The buffer used for output
     private let buffer = BufferList()


ianpartridge:
I'm not against making this change, but would an alternative be that we add another `userInfo` dictionary to `RouterResponse`? Then your middleware could stash the base context in the response instead, and your code would be:

```
extension RouterResponse {
    @discardableResult
    func render(_ resource: String, viewModel: ViewModel) throws -> RouterResponse {
        return try self.render(resource, context: self.userInfo.merging(viewModel.contents))
    }
}
```

Just a thought.

ianpartridge:
OK, we're happy to add a `userInfo` dictionary to RouterResponse. It provides nice symmetry with RouterRequest and keeps the two types isolated from each other (it's really an implementation detail that RouterResponse even knows about RouterRequest).

I'm going to close this PR - would you be willing to open another that adds the new dictionary to RouterResponse?

svanimpe:
@ianpartridge Is this simply a matter of adding the property? Or are there complications?

ianpartridge:
Functionally it's just adding the new public property to RouterResponse, along with a suitable doc comment like the one on RouterRequest.userInfo.

It would be nice to have some tests too. These could cover your usecase of populating the dictionary via a middleware, then reading the data out in a route handler and/or RouterResponse extension.  The existing RouterResponse tests are at https://github.com/IBM-Swift/Kitura/blob/master/Tests/KituraTests/TestResponse.swift and show an examples of a DummyErrorMiddleware which would be an easy pattern to follow.

Issue #1249, opened 2018-04-14
CodableRouter: unprocessableEntity should return the error to client


hactar:
## Context and Description
CodableRouter decodes JSON into a Codable object. If a client posts/gets/whatever a JSON object missing a required parameter, the CodableRouter correctly returns a 422 unprocessable entity and the current unreleased version even logs that to the server.

The person coding the client may not have access to the server logs however, and so it would be great if kitura would actually output the decoding error to the client as this allows debugging any malformed/incorrect json without the aid of the server developer.

This is especially important if you are developing a public api and want to help 3rd party developers with why their requests aren't working - these will never have access to the server logs.

### Environment Details
Swift 4.0.3, Kitura 2.2.x

### Steps to Reproduce

1) Create a simple kitura server app that accepts a codable via post route, such as:

```
public struct SuggestAction: Codable {
    public let title: String
    public let value: String
}
```

2) Submit a post to the server that isn't correct, for example its missing the "title":

```
{
  value: "woof"
}
```

### Expected vs. Actual Behaviour
- Expected to happen...
I should get a 422 http status, and an error message sent back to the client with the decoding error, such as:
`keyNotFound(CodingKeys(stringValue: "title", intValue: nil), Swift.DecodingError.Context(codingPath: [], debugDescription: "No value associated with key CodingKeys(stringValue: \"title\", intValue: nil) (\"title\").", underlyingError: nil))`
- Actually happened...
I only get a 422 http status back, and if I have no access to the server logs, I can only guess as to why the server did not accept my post.

### Additional
- I do not have a proposal of how to actually format the returned error. Should kitura just return the raw error string from the decode catch? should it return it in some JSON format? Personally I wouldn't care, as long as the error is returned in some form.
- Maybe make this toggable incase someone prefers the "quiet" option that kitura currently has.

### How to fix
Once the Kitura Team agrees upon a format of returning the error, all it needs is a response.send here:
https://github.com/IBM-Swift/Kitura/blob/661fc8a169b1718f4649aa990d9e745a01663c93/Sources/Kitura/CodableRouter.swift#L730
and maybe here:
https://github.com/IBM-Swift/Kitura/blob/661fc8a169b1718f4649aa990d9e745a01663c93/Sources/Kitura/CodableRouter.swift#L754

I added a `response.send("Decode Error: \(error)")` to my version. I'm sure this could be done prettier, but this at least gives client devs an idea of what went wrong.

ianpartridge:
[...]

PR #1256, opened 2018-04-21
Send JSON decoding error messages to clients in codable routes

hactar:
## Description
See #1249 
As usual Linux Swift and MacOS Swift behave differently, so both have been covered in this PR (see comments in files).
## Motivation and Context
See #1249 

## How Has This Been Tested?
Tests have been added and run. I have not used this PR in my own project yet as that project is still Kitura 2.2, but similar changes to Kitura 2.2 work fine for me.

## Checklist:


hactar:
@ianpartridge Your requested changes have been implemented. Please re-review.

Sources/Kitura/CodableRouter.swift
```diff
@@ -728,6 +728,12 @@ public struct CodableHelpers {
         } catch {
             Log.error("Failed to read Codable input from request: \(error)")
             response.status(.unprocessableEntity)
+            if let decodingError = error as? DecodingError {
+                response.send("Could not decode received JSON: \(decodingError.humanReadableDescription)")
+            } else {
+                // Linux Swift does not send a DecodingError when the JSON is invalid, instead it sends Error "The operation could not be completed"
+                response.send("Could not decode received JSON.")
+            }
```
djones6: Do we need to explicitly set `Content-Type` to `text/plain` here or is that the default?

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -179,11 +179,11 @@ class TestCodableRouter: KituraTest {
             
             .request("post", path: "/urlencoded", urlEncodedString: "encoding=valid&failed=match")
             .hasStatus(.unprocessableEntity)
-            .hasNoData()
+            .hasData()
```
djones6: We ought to be checking the `Content-Type` of the response is `text/plain`

Sources/Kitura/DecodingError+Extensions.swift
```diff
@@ -0,0 +1,56 @@
...
+            
+            var prefixString = DecodingError.codingKeyAsPrefixString(from: context.codingPath)
+            if prefixString.count > 0 {
+                prefixString = "\(prefixString)."
+            }
+            return "The required key '\(prefixString)\(type.stringValue)' not found."
+        case .dataCorrupted(let context):
+            // Linux does not get to this state but sends an Error "The operation could not be completed" instead. Future proofing this though just in case.
+            #if os(Linux)
+            // Linux wants a force downcast, MacOS doesn't.
```
djones6: I think this is okay.  If the casting of NSError on Linux changes in a future Swift release, we'll get a compilation error and have the usual process of guarding the Linux code based on swift version.

Tests/KituraTests/TestCodableRouter.swift
djones6: Actually, there's precedent for us sending plain text data without setting the Content-Type.  This is fine for now.

Tests/KituraTests/TestDecodingErrorExtension.swift
```diff
@@ -0,0 +1,150 @@
...
+"""
+        do {
+            _ = try JSONDecoder().decode(TestCodable.self, from: testJSON.data(using: .utf8)!)
+            XCTFail("We should have had a decoding error, but had no error.")
+        } catch  {
+            if let decodingError = error as? DecodingError {
+                XCTAssert(decodingError.humanReadableDescription.contains("The JSON appears to be malformed."), "DecodingError.humanReadableDescription not what we expected.")
+            } else {
+                // Linux Swift does not throw a DecodingError for malformed JSON, an Error "The operation could not be completed"
+                #if !os(Linux)
```
djones6: The only issue I can see here is that in the future, if Linux changes to be consistent with Darwin, we'll fail to make this assertion.  Since we don't know if/when this may be addressed, we can't add a Swift version check (well, we could, but it would be along the lines of `#if !os(Linux) || swift(>4.1)` and we'd have to bump the version each time there was another release that didn't fix it).  We should remember to revisit this in the future.

PR #1257, opened 2018-04-25
Raw if no content

Andrew-Lees11:
## Description
Changed the BodyParser to parse the body as raw data if no Content-Type header is present. This then makes the behavior match if the Content-Type provides is unknown.

Have updated the Docs to reflect this change and expanded on the usage Example.

Have added a test for this behaviour

## Motivation and Context
This was requested on the slack channel after a user got stuck trying to use the Body parser without having specified a Content-Type. 

## How Has This Been Tested?
- Test has been added and is passing
- Previous Kitura tests are also passing

## Checklist:

### DIFF
--- a/Sources/Kitura/bodyParser/BodyParser.swift
+++ b/Sources/Kitura/bodyParser/BodyParser.swift
@@ -28,7 +28,7 @@ import Foundation
 ///```swift
 ///router.all("/name", middleware: BodyParser())
 ///router.post("/name") { request, response, next in
-///    guard let parsedBody = request.body, let jsonBody = parsedBody.asJSON else {
+///    guard let jsonBody = request.body?.asJSON else {
 ///        next()
 ///        return
 ///    }
--- a/Sources/Kitura/bodyParser/BodyParser.swift
+++ b/Sources/Kitura/bodyParser/BodyParser.swift
@@ -21,12 +21,20 @@ import Foundation
 
 // MARK: BodyParser
 
-/// The `BodyParser` parses the body of the request prior to sending it to the handler. It reads the Content-Type of the message header and populates the `RouterRequest` body field with a `ParsedBody` enumeration (e.g. json, raw, text, urlEncoded).
+/// The `BodyParser` parses the body of the request prior to sending it to the handler. It reads the Content-Type of the message header and populates the `RouterRequest` body field with a corresponding `ParsedBody` enumeration ("application/json" -> JSON, "text/*" -> text, "application/x-www-form-urlencoded" -> URLEncoded, "multipart/*" -> multiPart, missing or any other Content-Type -> raw).
 /// In order for the BodyParser to be used it must first be registered with any routes that are interested in the ParsedBody payload.
 ///### Usage Example: ###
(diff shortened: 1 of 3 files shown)


Sources/Kitura/bodyParser/BodyParser.swift
```diff
@@ -21,12 +21,20 @@ import Foundation
...
+/// The `BodyParser` parses the body of the request prior to sending it to the handler. It reads the Content-Type of the message header and populates the `RouterRequest` body field with a corresponding `ParsedBody` enumeration ("application/json" -> JSON, "text/*" -> text, "application/x-www-form-urlencoded" -> URLEncoded, "multipart/*" -> multiPart, missing or any other Content-Type -> raw).
 /// In order for the BodyParser to be used it must first be registered with any routes that are interested in the ParsedBody payload.
 ///### Usage Example: ###
-/// In this example, all routes to the BodyParser middleware are registered to the `BodyParser` middleware.
+/// In this example, all routes to the BodyParser middleware are registered to the `BodyParser` middleware. An request with "contentType" = application/json is received. It is then parse as JSON and the value for "name" is returned in the response.
 ///```swift
-///   router.all("/*", middleware: BodyParser())
+///router.all("/name", middleware: BodyParser())
+///router.post("/name") { request, response, next in
+///    guard let parsedBody = request.body, let jsonBody = parsedBody.asJSON else {
```
ianpartridge: Can we do `guard let jsonBody = request.body?.asJSON else { }`?

Sources/Kitura/bodyParser/BodyParser.swift
```diff
@@ -60,11 +68,10 @@ public class BodyParser: RouterMiddleware {
             return next() // the body was already parsed
         }
 
-        guard request.headers["Content-Length"] != nil,
-            let contentType = request.headers["Content-Type"] else {
-                return next()
+        guard request.headers["Content-Length"] != nil else {
```
ianpartridge: Can't this just be `guard let contentType = request.headers["Content-Type"] else { }`?

Andrew-Lees11:
I have changed the test since they were not properly covering the use case. They will now fail since at some point before reaching my changed code the content-type header is being set as "application/x-www-form-urlencoded". I have been unable to find where this is happening with debugging.

I have tested the changes in this pull request with a running server and when sending a post without a content-type header, the content-type header was being set to text/plain and i was again unable to find where.

Sources/Kitura/bodyParser/BodyParser.swift
Andrew-Lees11: So if request.headers["Content-Type"] is allowed to be nil is can't be in a guard let. It have been moved down and made a normal let.
At the moment, request.headers["Content-Length"] is not allowed to be nil or we return. Would we want this header to also be allowed to be nil?

ianpartridge:
@Andrew-Lees11 this failed CI.

Andrew-Lees11:
I am closing this pull request since It is currently failing and I am not able to commit the time to fix this for a fairly minor change. May revisit it at a later date.

PR #1258, opened 2018-04-25
Add asRaw to ParsedBody

Andrew-Lees11:
## Description
Added an as raw helper function to parsed body. 

## Motivation and Context

This means instead of:

if case let .raw(data) = requestBody {}

You can do:

if let data = requestBody.asRaw {}

Which is more readable and now matched the other types (JSON, URLEncoded, text)

## How Has This Been Tested?
Test for bodyparser has been changed to use this helper function and is passing

## Checklist:


Andrew-Lees11:
The test change from case let .raw(data) to asRaw uses this tests this helper function alongside the asJSON, asText etc functions and should cover this change.

PR #1259, opened 2018-04-25
Add a userInfo property to RouterResponse

svanimpe:
This change adds a `userInfo` property to `RouterResponse`. The associated test checks if one handler can store data in `userInfo` and a subsequent handler can read it.

## Motivation and Context

This supports the use case of building a response in several steps. Handlers/middlewares can store intermediate data in `userInfo` so it can be used by the final handler to build the response.

## How Has This Been Tested?

A test scenario is included.

## Checklist:

### DIFF
--- a/Sources/Kitura/RouterResponse.swift
+++ b/Sources/Kitura/RouterResponse.swift
@@ -116,6 +116,9 @@ public class RouterResponse {
         }
     }
 
+    /// User info.
+    public var userInfo: [String: Any] = [:]
+    
     /// Initialize a `RouterResponse` instance
     ///
     /// - Parameter response: The `ServerResponse` object to work with
--- a/Sources/Kitura/RouterResponse.swift
+++ b/Sources/Kitura/RouterResponse.swift
@@ -117,6 +117,7 @@ public class RouterResponse {
     }
 
     /// User info.
+    /// Can be used by middlewares and handlers to store and pass information on to subsequent handlers.
     public var userInfo: [String: Any] = [:]
     
     /// Initialize a `RouterResponse` instance
--- a/Tests/KituraTests/TestResponse.swift
+++ b/Tests/KituraTests/TestResponse.swift
@@ -1087,6 +1087,22 @@ class TestResponse: KituraTest {
         })
     }
 
+    func testUserInfo() {
+        performServerTest(router) { expectation in
+            self.performRequest("get", path: "/user_info", callback: { response in
+                XCTAssertNotNil(response, "ERROR!!! ClientRequest response object was nil")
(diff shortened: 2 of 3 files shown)


Tests/KituraTests/TestResponse.swift
```diff
@@ -1481,6 +1498,22 @@ class TestResponse: KituraTest {
...
+        
+        router.get("user_info", handler: {
+            _, response, next in
+            // Store something in userInfo
+            response.userInfo["greeting"] = "hello"
+            next()
+        }, {
+            _, response, next in
+            // Read the value that should be stored in userInfo
+            guard let greeting = response.userInfo["greeting"] as? String else {
```
ianpartridge: This relies on the handlers being run by Kitura in the order that they are defined. Although we don't explicitly document this behaviour, it is the same as Express which also runs handlers in order. As such, we are unlikely to change this even though technically it is relying on undefined behaviour. A roundabout way of saying this is probably OK!

svanimpe: Ah OK, I didn't know this was undefined and assumed they were called in the order they are defined. Isn't this also why a global error handler needs to be registered last?

Sources/Kitura/RouterResponse.swift
```diff
@@ -116,6 +116,9 @@ public class RouterResponse {
         }
     }
 
+    /// User info.
```
djones6: Although I understand this is a copy of the equivalent comment on `RouterRequest`, I think it would be worth fleshing out this comment, perhaps with some guidance on when someone might want to use it and/or a usage example.

In particular I think it would be useful to disambiguate 'user' in this context from the identity of a remote client.

Tests/KituraTests/TestResponse.swift
djones6: @ianpartridge I was under the impression that we did rely on the handlers being invoked in definition order, for example to ensure that middlewares run (eg. session) before handlers that depend on them.  We should definitely document it.

Sources/Kitura/RouterResponse.swift
svanimpe: What about:

```swift
/// User info.
/// Can be used by middlewares and handlers to store and pass information on to subsequent handlers.
```

ianpartridge: Sounds good to me.  Would you mind adding this to the `RouterRequest` dictionary too, in this PR?

svanimpe:
@ianpartridge @djones6 The comments have been changed as requested.

ianpartridge:
LGTM - @djones6 please can you take another look.

djones6:
The comment seems fine - thanks.

PR #1260, opened 2018-04-26
update bodyparser docs example

Andrew-Lees11:
## Description
Updated the Jazzy docs for Body parser to say which content type headers match to ParsedBody types.
Have also added a usage example for getting a JSON body

## Motivation and Context
Requested by a user on slack after he was getting nil from the body parser when he wasn't sending a content type header.

## How Has This Been Tested?
Only a docs change so no added tests

### DIFF
--- a/Sources/Kitura/bodyParser/BodyParser.swift
+++ b/Sources/Kitura/bodyParser/BodyParser.swift
@@ -20,33 +20,34 @@ import LoggerAPI
 import Foundation
 
 // MARK: BodyParser
-
-/// The `BodyParser` parses the body of the request prior to sending it to the handler. It reads the Content-Type of the message header and populates the `RouterRequest` body field with a corresponding `ParsedBody` enumeration.
-/// In order for the BodyParser to be used it must first be registered with any routes that are interested in the ParsedBody payload.
-/// ### ParsedBody enumerations: ###
-/// ```swift
-/// "application/json" -> JSON: [String: Any]
-/// "text/*" -> text: String
-/// "application/x-www-form-urlencoded" -> URLEncoded: [String:String]
-/// "multipart/*" -> multiPart: [`Part`]
-/// ```
-/// If you provide any other Content-Type header, `ParsedBody` will be the raw bytes as a `Data` object.
-///
-/// If you have not declared a content type, `ParsedBody` will be `nil`
-/// ### Usage Example: ###
-/// In this example, all routes to the BodyParser middleware are registered to the `BodyParser` middleware. A request with "contentType" == "application/json" is received. It is then parsed as JSON and the value for "name" is returned in the response.
-/// ```swift
-/// router.all("/name", middleware: BodyParser())
-/// router.post("/name") { request, response, next in
-///     guard let jsonBody = request.body?.asJSON else {
-///         next()
-///         return
-///     }
-///     let name = jsonBody["name"] as? String ?? ""
-///     try response.send("Hello \(name)").end()
-/// }
-/// ```
-/// __Note__: When using Codable Routing in Kitura 2.x the BodyParser should not be registered to any codable routes (doing so will log the following error "No data in request. Codable routes do not allow the use of a BodyParser." and the route handler will not be executed).
+/**
+ The `BodyParser` parses the body of the request prior to sending it to the handler. It reads the Content-Type of the message header and populates the `RouterRequest` body field with a corresponding `ParsedBody` enumeration.  Each enumerations has a helper function, e.g. `asURLEncoded`, for accessing the data.
+ In order for the BodyParser to be used it must first be registered with any routes that are interested in the `ParsedBody` payload.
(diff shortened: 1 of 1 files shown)


Sources/Kitura/bodyParser/BodyParser.swift
```diff
@@ -21,13 +21,20 @@ import Foundation
...
 
-/// The `BodyParser` parses the body of the request prior to sending it to the handler. It reads the Content-Type of the message header and populates the `RouterRequest` body field with a `ParsedBody` enumeration (e.g. json, raw, text, urlEncoded).
+/// The `BodyParser` parses the body of the request prior to sending it to the handler. It reads the Content-Type of the message header and populates the `RouterRequest` body field with a corresponding `ParsedBody` enumeration ("application/json" -> JSON, "text/*" -> text, "application/x-www-form-urlencoded" -> URLEncoded, "multipart/*" -> multiPart, no declared content type -> nil, any other Content-Type -> raw).
 /// In order for the BodyParser to be used it must first be registered with any routes that are interested in the ParsedBody payload.
 ///### Usage Example: ###
-/// In this example, all routes to the BodyParser middleware are registered to the `BodyParser` middleware.
+/// In this example, all routes to the BodyParser middleware are registered to the `BodyParser` middleware. An request with "contentType" = application/json is received. It is then parse as JSON and the value for "name" is returned in the response.
```
ianpartridge: An request => A request
parse => parsed

Sources/Kitura/bodyParser/BodyParser.swift
```diff
@@ -21,13 +21,20 @@ import Foundation
...
-///```
+///router.all("/name", middleware: BodyParser())
+///router.post("/name") { request, response, next in
+///    guard let jsonBody = request.body?.asJSON else {
+///        next()
+///        return
+///    }
+///    let name = jsonBody["name"] as? String ?? ""
+///    try response.send("Hello \(name)").end()
+///}
```
helenmasters: The formatting here has broken the Note below so that it displays as:
"__Note__: When using Codable Routing in Kitura 2..." within the Usage Example.

Sources/Kitura/bodyParser/BodyParser.swift
```diff
@@ -21,13 +21,20 @@ import Foundation
 
 // MARK: BodyParser
 
-/// The `BodyParser` parses the body of the request prior to sending it to the handler. It reads the Content-Type of the message header and populates the `RouterRequest` body field with a `ParsedBody` enumeration (e.g. json, raw, text, urlEncoded).
+/// The `BodyParser` parses the body of the request prior to sending it to the handler. It reads the Content-Type of the message header and populates the `RouterRequest` body field with a corresponding `ParsedBody` enumeration ("application/json" -> JSON, "text/*" -> text, "application/x-www-form-urlencoded" -> URLEncoded, "multipart/*" -> multiPart, no declared content type -> nil, any other Content-Type -> raw).
```
helenmasters: This is not at all easy to read in this format.

I've spoken with David Jones and we suggest removing the following wording:
("application/json" -> JSON, "text/*" -> text, "application/x-www-form-urlencoded" -> URLEncoded, "multipart/*" -> multiPart, no declared content type -> nil, any other Content-Type -> raw).

And instead add a new heading of the same font size as Usage Example but called:
ParsedBody Enumeration
This can then describe the ParsedBody enumeration for each case
e.g. "application/json" to JSON

The last two entries should be described as sentences outside the table (I think). e.g. If the message header has no declared content type the ParsedBody enumeration will contain `nil`. If the message header contains any other content type the ParserdBody enumeration will be.....?

helenmasters:
@Andrew-Lees11 please can you send this to @djones6 to review when you've made the requested changes, thanks

Sources/Kitura/bodyParser/BodyParser.swift
```diff
@@ -21,13 +21,31 @@ import Foundation
...
-/// In this example, all routes to the BodyParser middleware are registered to the `BodyParser` middleware.
-///```swift
-///   router.all("/*", middleware: BodyParser())
-///```
+/// ### ParsedBody enumerations: ###
+/// ```swift
+/// "application/json" -> JSON: [String: Any]
+/// "text/*" -> text: String
+/// "application/x-www-form-urlencoded" -> URLEncoded: [String:String]
+/// "multipart/*" -> multiPart: [`Part`]
```
djones6: I find this formatting confusing - I think what we're trying to convey here is the mapping from the incoming `Content-Type` to our internal representation of the body data.  How about an opening statement such as:
> This represents the mapping from the incoming `Content-Type` to an internal representation of the body data, as follows:
> - "application/json": `.json([String: Any])`
> - "text/*": `.text(String)`
> - ...etc
>
> Any other content type maps to the `.raw(Data)` case.
>
> For each case, there is a corresponding convenience property for accessing the associated data - for example, `asJSON` returns a `[String: Any]?`.

Issue #1261, opened 2018-04-30
Codable forms: parsing non-String values


svanimpe:
## Context and Description

I am trying to replace manual form parsing with Codable forms. However, the behavior when dealing with empty values is not as I would expect it.

### Environment Details

Kitura 2.3
Xcode 9.3

### Steps to Reproduce

Codable type:

```
struct LocationForm: Codable {
    let address: String?
    let city: String?
    let country: String?
    let latitude: Double?
    let longitude: Double?
}
```

Form sent: `address=&city=&country=&latitude=&longitude=`

These are all `<input>`s for which a value was never set.

### Expected vs. Actual Behaviour

- [x] Expected to happen: `latitude` and `longitude` are `nil`.

- [x] Actually happened: An error is thrown:

```
[2018-04-30T08:53:50.270+02:00] [ERROR] [QueryDecoder.swift:145 decodingError()] Could not process field named 'latitude'.
[2018-04-30T08:53:50.273+02:00] [ERROR] [RouterMiddlewareWalker.swift:72 next()] dataCorrupted(Swift.DecodingError.Context(codingPath: [CodingKeys(stringValue: "latitude", intValue: nil)], debugDescription: "Could not process field named \'latitude\'.", underlyingError: nil))
```

I would expect non-String values to be `nil` if either the value isn't provided, or it cannot be parsed to the correct type. This is in line with code Codable forms are supposed to replace, as in the example on https://developer.ibm.com/swift/2018/04/17/announcing-kitura-2-3/, where `id` should be an Int:

```
router.post("/form") { request, response, next in
    guard let encodedBody = request.body,
        case .urlEncoded(let body) = encodedBody,
        let idString = body["id"],
        let id = Int(idString),
        let name = body["name"]
    else {
        return try response.status(.unprocessableEntity).end()
    }
    let user = User(id: id, name: name)
    // Application logic
    response.status(.created).send(json: user)
    next()
}
```

Andrew-Lees11:
Hello Steven

Thank you for raising this issue and providing so much detail.

When we receive a String such as "address=&city=&country=&latitude=&longitude=" we process the = as being an empty string "". Since longitude and latitude are supposed to be doubles this causes the unprocessableEntity error to be thrown. 

To get a nil value for an optional field using the Query decoder the inputs need to not be there. E.g:
address=&city=&country=

Would set address, city, country to be empty string and latitude, longitude would be nil. 

[...]

PR #1262, opened 2018-04-30
Fix: URLEncoded forms treat empty string as nil

Andrew-Lees11:
## Description
This pull requests changes the read as function for URLEncoded forms to treat the empty string as nil. 

## Motivation and Context
In response to issue #1261. When using HTML forms to send data, An empty form will send an empty string to signify nil for all data types (e.g. numbers, dates). 

When decoding a form with optional inputs for these data types the user would get unprocessableEntity since the empty string cannot be converted to a Bool/ date etc. 

This change will also mean that a form submitting an empty string will treat it as being nil instead of the empty string and it will be rejected if the field is non optional.

## How Has This Been Tested?
Tests have been added to check for a number submitted as an empty string being accepted for a struct with an optional value. Test have also been added to show an empty string being rejected for a non optional String and number.

## Checklist:

### DIFF
--- a/Sources/Kitura/String+Extensions.swift
+++ b/Sources/Kitura/String+Extensions.swift
@@ -26,7 +26,7 @@ extension String {
         var result: [String: String] = [:]
         for item in self.components(separatedBy: "&") {
             let (key, value) = item.keyAndDecodedValue
-            if let value = value {
+            if let value = value, !value.isEmpty {
                 // If value already exists for this key, append it
                 if let existingValue = result[key] {
                     result[key] = "\(existingValue),\(value)"
@@ -46,7 +46,7 @@ extension String {
 
         for item in self.components(separatedBy: "&") {
             let (key, value) = item.keyAndDecodedValue
-            if let value = value {
+            if let value = value, !value.isEmpty {
                 result[key, default: []].append(value)
             }
         }
--- a/Tests/KituraTests/TestCodableRouter.swift
+++ b/Tests/KituraTests/TestCodableRouter.swift
@@ -152,8 +152,13 @@ class TestCodableRouter: KituraTest {
             print("POST on /urlencoded for user \(user)")
             respondWith(user, nil)
         }
+        router.post("/optionalurlencoded") { (OptionalUser: OptionalUser, respondWith: (OptionalUser?, RequestError?) -> Void) in
+            print("POST on /urlencoded for user \(OptionalUser)")
+            respondWith(OptionalUser, nil)
+        }
 
         let user = User(id: 4, name: "David")
+        let optionalUser = OptionalUser(id: nil, name: "David")
         buildServerTest(router, timeout: 30)
             .request("post", path: "/users", data: user)
             .hasStatus(.created)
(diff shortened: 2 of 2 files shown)


ianpartridge:
@svanimpe

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -150,8 +150,13 @@ class TestCodableRouter: KituraTest {
             print("POST on /urlencoded for user \(user)")
             respondWith(user, nil)
         }
+        router.post("/optionalurlencoded") { (OptionalUser: OptionalUser, respondWith: (OptionalUser?, RequestError?) -> Void) in
+            print("POST on /urlencoded for user \(OptionalUser)")
+            respondWith(OptionalUser, nil)
+        }
 
         let user = User(id: 4, name: "David")
+        let optionalUser = OptionalUser(id: nil, name: "David")
```
EnriqueL8: What happens if we have this User struct: 
```
struct User: Codable {
  var id: Int?
  var name: String
}
```
and we get `id=4&name=`, we throw an error because `name` cannot be `nil`?

Andrew-Lees11: Yes. It will treat the empty string as nil even for strings. So if you have a non-optional String variable and you get the empty string it will throw an error.

Sources/Kitura/RouterRequest.swift
```diff
@@ -188,7 +188,8 @@ public class RouterRequest {
             guard let urlKeyValuePairs = body?.urlDecodedFieldValuePairs else {
                 throw Error.failedToParseRequestBody(body: body ?? "Failed to read body as String")
             }
-            return try QueryDecoder(dictionary: urlKeyValuePairs).decode(type)
+            let removedEmptyStrings = urlKeyValuePairs.filter { $0.value != "" }
```
ianpartridge: Would it be better to use `isEmpty` instead?

ianpartridge: Why don't we just change `urlDecodedFieldValuePairs` so it's `[String: String?]` instead?

Andrew-Lees11:
The check for query String being nil has been moved inside urlDecodedFieldValuePairs in String+Extensions.swift. This means the change will also apply to standard query parameters so when sending a form with get it will also treat the empty string as nil.

Andrew-Lees11:
Corresponding change for Kitura-net is [here](https://github.com/IBM-Swift/Kitura-net/pull/267). These should probably be marged as minor tags together so they remain consistent.

Sources/Kitura/String+Extensions.swift
```diff
@@ -32,20 +32,20 @@ extension String {
 
             let key = String(item[..<range.lowerBound])
             let value = String(item[range.upperBound...])
-
-            let valueReplacingPlus = value.replacingOccurrences(of: "+", with: " ")
-            if let decodedValue = valueReplacingPlus.removingPercentEncoding {
-                if let value = result[key] {
-                    result[key] = "\(value),\(decodedValue)"
+            if value != "" {
```
ianpartridge: I think we can improve the flow...

```swift
let value = String(item[range.upperBound...])
if value.isEmpty {
    continue
}
let valueReplacingPlus = ...
```

ianpartridge: And we don't need to calculate the `key` until after we've validated the value.

Andrew-Lees11:
Closing in favour of new [PR](https://github.com/IBM-Swift/KituraContracts/pull/21/files) changing behavior inside Kitura Contracts.

Andrew-Lees11:
@svanimpe It will mean the raw parameters will still show as param= but if you use the decoder it will decode as nil.

PR #1281, opened 2018-05-29
generate swagger

nhardman:
Support for Swagger document generation from Kitura

## Description
This PR implements code that intercepts the registration of codable routes and builds an in-memory model of the routes, but structured in the same way as an OpenAPI (Swagger) 2.0 API document. To use this, there is an externalized property (swaggerJSON) on the router, which can be read. This will serialize the data into a JSON string.

## Motivation and Context
This was needed to fulfill the requirement for a more complete OpenAPI support. Using this document will allow server APIs to be replicated and consumed easily, allowing for the rapid generation of servers and testing of conformance to APIs.

## How Has This Been Tested?
Swift test was used to implement tests for the generation of the expected data.
Hand testing was also used during development. 
Impact on other areas is minimal, although there would be a slight performance hit during route registration and serialization.

## Checklist:

- [ x ] If applicable, I have added tests to cover my changes.


Sources/Kitura/Router.swift
```diff
@@ -73,6 +73,22 @@ public class Router {
     /// and used to handle request's url parameters.
     fileprivate var parameterHandlers = [String : [RouterParameterHandler]]()
 
+    /// Contains the structures needed for swagger document generation
+    var swagger: SwaggerDocument
+
+    /// Computed property
```
ianpartridge: Remove this.

Sources/Kitura/Router.swift
```diff
@@ -73,6 +73,22 @@ public class Router {
...
+    ///
+    /// returns the current in-memory representation of the API as a json string.
```
ianpartridge: Returns the current in-memory representation of Codable routes as a Swagger document in JSON format, or nil if the document cannot be generated.

Sources/Kitura/Router.swift
```diff
@@ -73,6 +73,22 @@ public class Router {
...
+    public var swaggerJSON: String? {
+        get {
```
ianpartridge: Remove the `get { }`, it is only required if you have a `set { }` as well.

```swift
public var swaggerJSON: String? {
    do {
        return try self.swagger.serializeAPI(format: .json)
    } catch {
        return nil
    }
}
```

Sources/Kitura/Router.swift
```diff
@@ -102,7 +118,12 @@ public class Router {
     /// ```
     /// - Parameter mergeParameters: Optional parameter to specify if the router should be able to access parameters
     ///                                 from its parent router. Defaults to `false` if not specified.
+    /// Initialize a `Router` instance
```
ianpartridge: These comments seem to be duplicated from above, please remove.

Sources/Kitura/Router.swift
nhardman: done

Sources/Kitura/Router.swift
nhardman: done

Sources/Kitura/Router.swift
nhardman: done

Sources/Kitura/Router.swift
nhardman: done

PR #1283, opened 2018-05-30
fix: remove trailing Whitespace

Andrew-Lees11:
## Description
Removed the trailing white-space and extra lines in the TypeSafeMiddleware tests.
This will remove the warnings from the travis build and neaten up the code a bit.

### DIFF
--- a/Tests/KituraTests/TestTypeSafeMiddleware.swift
+++ b/Tests/KituraTests/TestTypeSafeMiddleware.swift
@@ -1,0 +1,919 @@
+/**
+ * Copyright IBM Corporation 2018
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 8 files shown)


Tests/KituraTests/TestTypeSafeMiddleware.swift
```diff
@@ -48,7 +48,7 @@ class TestTypeSafeMiddleware: KituraTest {
             ("testSingleMiddlewarePut", testSingleMiddlewarePut),
             ("testMultipleMiddlewarePut", testMultipleMiddlewarePut),
             ("testSingleMiddlewarePatch", testSingleMiddlewarePatch),
-            ("testMultipleMiddlewarePatch", testMultipleMiddlewarePatch),
```
djones6: Is there a specific "disallow commas at the end of lists" rule? If so, I'd consider disabling it rather than making this change.

Specifically, I like the ability to include the same delimiters on all lines of the list (particularly when it is laid out with the opening and closing braces on separate lines, as is the case here).  It means that adding new entries is a simpler operation, and if adding to the end of the list, produces a cleaner diff.

PR #1284, opened 2018-05-30
feat: Add optional QueryParams in closure for  get and delete

EnriqueL8:
## Description
This PR enables optional QueryParams structure in `get` and `delete` routes.

## Motivation and Context

When declaring a route with QueryParameters, such as: 
```
router.get("/query") { (query: MyQuery, respondWith: @escaping ([User]?, RequestError?) -> Void) in
```
Currently, we cannot make the query optional, and we would have to declare another route.
This issue resolves this by adding a new closure to the **get** and **delete** accepting an optional QueryParams, so we could write:
```
router.get("/query") { (query: MyQuery?, respondWith: @escaping ([User]?, RequestError?) -> Void) in
```

This route would accept a **get** call on `/query` and `/query?age=5`.

## How Has This Been Tested?

Test have been added.

## Checklist:

### DIFF
--- a/Sources/Kitura/CodableRouter.swift
+++ b/Sources/Kitura/CodableRouter.swift
@@ -481,6 +481,7 @@ extension Router {
 
     // Get w/Optional Query Parameters
     fileprivate func getSafely<Q: QueryParams, O: Codable>(_ route: String, handler: @escaping (Q?, @escaping CodableArrayResultClosure<O>) -> Void) {
+        registerGetRoute(route: route, id: false, outputtype: O.self)
         get(route) { request, response, next in
             Log.verbose("Received GET (plural) type-safe request with Query Parameters")
             Log.verbose("Query Parameters: \(request.queryParameters)")
@@ -488,7 +489,7 @@ extension Router {
                 var query: Q? = nil
                 let queryParameters = request.queryParameters
                 if queryParameters.count > 0 {
-                  query = try QueryDecoder(dictionary: request.queryParameters).decode(Q.self)
+                    query = try QueryDecoder(dictionary: queryParameters).decode(Q.self)
                 }
                 handler(query, CodableHelpers.constructOutResultHandler(response: response, completion: next))
             } catch {
@@ -501,6 +502,7 @@ extension Router {
 
     // Get w/Optional Query Parameters with CodableResultClosure
     fileprivate func getSafely<Q: QueryParams, O: Codable>(_ route: String, handler: @escaping (Q?, @escaping CodableResultClosure<O>) -> Void) {
+        registerGetRoute(route: route, id: false, outputtype: O.self)
         get(route) { request, response, next in
             Log.verbose("Received GET (singular) type-safe request with Query Parameters")
             Log.verbose("Query Parameters: \(request.queryParameters)")
@@ -509,7 +511,7 @@ extension Router {
                 var query: Q? = nil
                 let queryParameters = request.queryParameters
                 if queryParameters.count > 0 {
-                  query = try QueryDecoder(dictionary: request.queryParameters).decode(Q.self)
+                    query = try QueryDecoder(dictionary: queryParameters).decode(Q.self)
                 }
                 handler(query, CodableHelpers.constructOutResultHandler(response: response, completion: next))
             } catch {
@@ -579,6 +581,7 @@ extension Router {
 
     // DELETE w/Optional Query Parameters
     fileprivate func deleteSafely<Q: QueryParams>(_ route: String, handler: @escaping (Q?, @escaping ResultClosure) -> Void) {
+        registerDeleteRoute(route: route, id: false)
         delete(route) { request, response, next in
             Log.verbose("Received DELETE type-safe request with Query Parameters")
             Log.verbose("Query Parameters: \(request.queryParameters)")
@@ -586,7 +589,7 @@ extension Router {
                 var query: Q? = nil
                 let queryParameters = request.queryParameters
                 if queryParameters.count > 0 {
-                  query = try QueryDecoder(dictionary: request.queryParameters).decode(Q.self)
+                    query = try QueryDecoder(dictionary: queryParameters).decode(Q.self)
                 }
                 handler(query, CodableHelpers.constructResultHandler(response: response, completion: next))
             } catch {
--- a/Sources/Kitura/CodableRouter.swift
+++ b/Sources/Kitura/CodableRouter.swift
@@ -152,6 +152,48 @@ extension Router {
     }
 
     /**
+     Setup a (QueryParams?, CodableArrayResultClosure) -> Void on the provided route which will be invoked when a request comes to the server.
+
+     ### Usage Example: ###
+     ````
+     // MyQuery is a codable struct defining the supported query parameters
+     // User is a struct object that conforms to Codable
+     router.get("/query") { (query: MyQuery?, respondWith: ([User]?, RequestError?) -> Void) in
+
+        ...
+
+        respondWith(users, nil)
+     }
+     ````
+     - Parameter route: A String specifying the pattern that needs to be matched, in order for the handler to be invoked.
+     - Parameter handler: A (QueryParams?, CodableArrayResultClosure) -> Void that gets invoked when a request comes to the server.
+     */
+    public func get<Q: QueryParams, O: Codable>(_ route: String, handler: @escaping (Q?, @escaping CodableArrayResultClosure<O>) -> Void) {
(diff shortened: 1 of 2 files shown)


Sources/Kitura/CodableRouter.swift
```diff
@@ -408,6 +470,46 @@ extension Router {
...
+    // Get w/Optional Query Parameters
+    fileprivate func getSafely<Q: QueryParams, O: Codable>(_ route: String, handler: @escaping (Q?, @escaping CodableArrayResultClosure<O>) -> Void) {
+        get(route) { request, response, next in
+            Log.verbose("Received GET (plural) type-safe request with Query Parameters")
+            Log.verbose("Query Parameters: \(request.queryParameters)")
+            do {
+                var query: Q? = nil
+                let queryParameters = request.queryParameters
+                if queryParameters.count > 0 {
+                  query = try QueryDecoder(dictionary: request.queryParameters).decode(Q.self)
```
ianpartridge: Indentation is off.

ianpartridge: Also, you can just do `dictionary: queryParameters`

Sources/Kitura/CodableRouter.swift
```diff
@@ -408,6 +470,46 @@ extension Router {
...
+    // Get w/Optional Query Parameters with CodableResultClosure
+    fileprivate func getSafely<Q: QueryParams, O: Codable>(_ route: String, handler: @escaping (Q?, @escaping CodableResultClosure<O>) -> Void) {
+        get(route) { request, response, next in
+            Log.verbose("Received GET (singular) type-safe request with Query Parameters")
+            Log.verbose("Query Parameters: \(request.queryParameters)")
+            // Define result handler
+            do {
+                var query: Q? = nil
+                let queryParameters = request.queryParameters
+                if queryParameters.count > 0 {
```
ianpartridge: Same comments as above.

Sources/Kitura/CodableRouter.swift
```diff
@@ -462,6 +564,26 @@ extension Router {
...
+    // DELETE w/Optional Query Parameters
+    fileprivate func deleteSafely<Q: QueryParams>(_ route: String, handler: @escaping (Q?, @escaping ResultClosure) -> Void) {
+        delete(route) { request, response, next in
+            Log.verbose("Received DELETE type-safe request with Query Parameters")
+            Log.verbose("Query Parameters: \(request.queryParameters)")
+            do {
+                var query: Q? = nil
+                let queryParameters = request.queryParameters
+                if queryParameters.count > 0 {
+                  query = try QueryDecoder(dictionary: request.queryParameters).decode(Q.self)
```
ianpartridge: Same comments as above.

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -716,6 +716,15 @@ class TestCodableRouter: KituraTest {
             respondWith(query, nil)
         }
 
+        router.get("/optionalquery") { (query: MyQuery?, respondWith: (MyQuery?, RequestError?) -> Void) in
+            if let query = query {
+              XCTAssertEqual(query, expectedQuery)
```
ianpartridge: Indentation.

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -746,16 +768,38 @@ class TestCodableRouter: KituraTest {
             respondWith([query], nil)
         }
 
+        router.get("/optionalquery") { (query: MyQuery?, respondWith: ([MyQuery]?, RequestError?) -> Void) in
+            if let query = query {
+              XCTAssertEqual(query, expectedQuery)
```
ianpartridge: Indentation.

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -776,6 +820,13 @@ class TestCodableRouter: KituraTest {
             respondWith(nil)
         }
 
+        router.delete("/optionalquery") { (query: MyQuery?, respondWith: (RequestError?) -> Void) in
+            if let query = query {
+              XCTAssertEqual(query, expectedQuery)
```
ianpartridge: Indentation.

Sources/Kitura/CodableRouter.swift
EnriqueL8: done

Sources/Kitura/CodableRouter.swift
EnriqueL8: done

Tests/KituraTests/TestCodableRouter.swift
EnriqueL8: done

Sources/Kitura/CodableRouter.swift
```diff
@@ -408,6 +470,46 @@ extension Router {
         }
     }
 
+    // Get w/Optional Query Parameters
+    fileprivate func getSafely<Q: QueryParams, O: Codable>(_ route: String, handler: @escaping (Q?, @escaping CodableArrayResultClosure<O>) -> Void) {
+        get(route) { request, response, next in
```
ianpartridge: Now that #1281 is merged, this needs a call to `registerGetRoute()` at the start of the function.

Sources/Kitura/CodableRouter.swift
```diff
@@ -408,6 +470,46 @@ extension Router {
...
+                // Http 400 error
+                response.status(.badRequest)
+                next()
+            }
+        }
+    }
+
+    // Get w/Optional Query Parameters with CodableResultClosure
+    fileprivate func getSafely<Q: QueryParams, O: Codable>(_ route: String, handler: @escaping (Q?, @escaping CodableResultClosure<O>) -> Void) {
+        get(route) { request, response, next in
```
ianpartridge: Now that #1281 is merged, this needs a call to `registerGetRoute()` at the start of the function.

Sources/Kitura/CodableRouter.swift
```diff
@@ -462,6 +564,26 @@ extension Router {
         }
     }
 
+    // DELETE w/Optional Query Parameters
+    fileprivate func deleteSafely<Q: QueryParams>(_ route: String, handler: @escaping (Q?, @escaping ResultClosure) -> Void) {
+        delete(route) { request, response, next in
```
ianpartridge: Now that #1281 is merged, this needs a call to `registerDeleteRoute()` at the start of the function.

Sources/Kitura/CodableRouter.swift
```diff
@@ -417,6 +479,48 @@ extension Router {
...
+        get(route) { request, response, next in
+            Log.verbose("Received GET (plural) type-safe request with Query Parameters")
+            Log.verbose("Query Parameters: \(request.queryParameters)")
+            do {
+                var query: Q? = nil
+                let queryParameters = request.queryParameters
+                if queryParameters.count > 0 {
+                    query = try QueryDecoder(dictionary: queryParameters).decode(Q.self)
+                }
+                handler(query, CodableHelpers.constructOutResultHandler(response: response, completion: next))
```
djones6: Would it be worth extracting the building of the query object into a helper?  Something like
```swift
func decodeQueryParams<Q: QueryParams>(_ request: RouterRequest) throws -> Q?
```
That way if we need more permutations of route patterns that accept `QueryParams?` there is less code to duplicate.

Ideally we could go a step further by avoiding the need for the do/catch, but that might take some more thinking (perhaps the helper could take a closure to invoke on success, or set the response status on failure).

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -726,6 +735,19 @@ class TestCodableRouter: KituraTest {
...
+            .hasContentType(withPrefix: "application/json")
+            .hasData(expectedQuery)
+
+            .request("get", path: "/optionalquery")
+            .hasStatus(.OK)
+            .hasNoData()
+
+            .request("get", path: "/optionalquery?param=badRequest")
+            .hasStatus(.badRequest)
+            .hasNoData()
```
djones6: I feel there are too many things being tested in one test function.  I appreciate there's precedent, but I'd prefer a separate test such as `testCodableGetSingleOptionalQueryParameters()`.

Because these tests make their assertions within the KituraTestBuilder functions, when one fails you only get the failing line within the helper (eg. `hasStatus`) function, and not the line of the `testXyz()` function that called it.  If each test only has one instance of `hasStatus(.OK)` then it's easy to tell what failed, but in this instance we are making the same assertion in 3 places.

PR #1290, opened 2018-05-31
Swagger: handle models with only optional properties

ianpartridge:
Currently, the Swagger generation does not work if a model only contains optional properties.

For example:

```swift
public struct Foo {
  public let id: Int?
}
 
router.post("/", handler: postHandler)
 
func postHandler(foo: Foo, completion: (Foo?, RequestError?) -> Void ) {
     completion(foo, nil)
 }
```

This fails because the model serialization code defines `required` as an optional array, and if it's nil the serialization fails.

This PR reworks the code so `required` is non-optional, and uses the empty array if there are no required properties.  Then, check if the array is empty before attempting to serialize it, because serializing an empty `required` array is not valid Swagger.


ianpartridge:
NB. This needs a test adding.

PR #1294, opened 2018-06-04
id type now set to string or integer

nhardman:
id parameter was not being set correctly. I have agreed with Enrique that the valid values for this will be either "string" or "integer". This fix picks the value from the registration of the codable route.


djones6:
This was superseded by #1316

PR #1297, opened 2018-06-06
feat: Public TypeAliases

EnriqueL8:
## Description

Adding typealias for KituraContracts types.

## Motivation and Context

For a user to use a type from KituraContracts, they would need to import KituraContacts in their application. Instead we expose these types in public typealiases in Kitura so that when they import Kitura they can access them.

## How Has This Been Tested?

Added tests.

## Checklist:

### DIFF
--- a/Sources/Kitura/TypeAliases.swift
+++ b/Sources/Kitura/TypeAliases.swift
@@ -1,5 +1,5 @@
 /*
- * Copyright IBM Corporation 2017
+ * Copyright IBM Corporation 2018
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


Sources/Kitura/TypeAliases.swift
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright IBM Corporation 2017
```
djones6: It's 2018

Tests/KituraTests/TestTypeAliases.swift
```diff
@@ -0,0 +1,53 @@
+/**
+ * Copyright IBM Corporation 2017
```
djones6: It's still 2018

Sources/Kitura/TypeAliases.swift
```diff
@@ -0,0 +1,49 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import KituraContracts
+
+/*
+  This file declares public typealiases for types stored in the dependencies.
+  The purpose is to expose these types at the top level without having to import a specific dependency.
+*/
```
djones6: Have you tested what Jazzy does with this file? Does it look correct? (I've had instances where successive comment blocks with no code separating them causes the documentation to vanish from the next code block).

Sources/Kitura/TypeAliases.swift
```diff
@@ -0,0 +1,49 @@
...
+
+/// Public TypeAlias for QueryParams Type from KituraContracts
```
djones6: I'd prefer we use the same technique of linking people through to the 'real' documentation as the existing Kitura typealiases, eg:
https://github.com/IBM-Swift/Kitura/blob/2.4.1/Sources/Kitura/CodableRouter.swift#L22

ianpartridge:
It turns out that Swift has an alternative way to do this. We could add this to Kitura:

```swift
@_exported import KituraContracts
```

Then, `KituraContracts` types could be used without a corresponding `import KituraContracts`. I'm closing this PR while we consider doing that.

PR #1299, opened 2018-06-13
update: Swift 4.1.2 as default and newer Swift 4.2 Travis builds

KyeMaloy97:
Adds a newer Swift 4.2 build to the Travis builds and sets the default Swift version from 4.1.1 to 4.1.2.


djones6:
We can't use a Swift 4.2 snapshot of 2018-06-05-a or later in Travis, because it requires Xcode 10 beta:
https://github.com/apple/swift/pull/17001

...and there is no Xcode 10 beta image yet:
https://github.com/travis-ci/travis-ci/issues/9730

PR #1304, opened 2018-06-29
Query params support for swagger

nhardman:
Decode any query params types that are used and build the appropriate 'query' parameter section for the route into a swagger document.

## How Has This Been Tested?
Unit tests have been written to test that the expected values of the query params are found. 

## Checklist:


seabaylea:
Two issues I've spotted:
1. This doesn't support optional QueryParams, eg:
```swift
    func getUsers(filter: UserQuery?, completion: @escaping ([User]?, RequestError?) -> Void) -> Void {
        User.findAll(matching: filter, completion)
    }
```
2. This doesn't handle the ORMs wrappers in the QueryParams, eg:
```swift
public struct UserQuery: QueryParams {
    let id: GreaterThan<Int>
}
```

ianpartridge:
I had a look at the OpenAPI spec and there does seem to be no way to specify that the entire query parameters are optional - the optionality is on the individual fields.  I guess we will have to make every field optional if the entire QueryParams is optional - not ideal but there you go.

Re: the ORM types - yes we should add this of course but I would much prefer we deliver this incrementally as the diff is already +1,247 −609.

djones6:
PRs for this have been merged (#1316, #1320, #1321, #1322).

PR #1308, opened 2018-07-17
[WIP]Allow users to opt for KituraNIO

pushkarnk:
Allow users to opt for KituraNIO while building their Kitura apps. 

## Description
KituraNIO support is currently available only on the `kitura-nio` branch. This pull request proposes introducing that support on the master branch and eventually making it generally available through a minor update. Users that desire to have the networking done via KituraNIO need to set up an env var and pass a compiler flag in the build step, in this way:
```
export KITURA_NIO=1 && swift build -Xswiftc -DKITURA_NIO
```
For users that desire to continue using KituraNet, there's no change required.

## Checklist:
- [ ] Add steps to travis.yml to use KituraNIO on trusty, xenial and macOS

### DIFF
--- a/Tests/KituraTests/KituraTest.swift
+++ b/Tests/KituraTests/KituraTest.swift
@@ -17,7 +17,7 @@
 import XCTest
 import Kitura
 
-#if NIO
+#if KITURA_NIO
 @testable import KituraNIO
 #else
 @testable import KituraNet
@@ -131,7 +131,7 @@ class KituraTest: XCTestCase {
 
         let server = HTTP.createServer()
 
-        #if NIO
+        #if KITURA_NIO
         server.allowPortReuse = true
         #endif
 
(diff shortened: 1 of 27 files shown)


Tests/KituraTests/KituraTest.swift
```diff
@@ -126,6 +130,11 @@ class KituraTest: XCTestCase {
         }
 
         let server = HTTP.createServer()
+
+        #if KITURA_NIO
+        server.allowPortReuse = true
```
djones6: Why do we need to enable port reuse here for NIO?

Tests/KituraTests/TestServer.swift
```diff
@@ -41,57 +46,87 @@ class TestServer: KituraTest {
 
     private func setupServerAndExpectations(expectStart: Bool, expectStop: Bool, expectFail: Bool, httpPort: Int?=nil, fastCgiPort: Int?=nil) {
         let router = Router()
+
+        #if KITURA_NIO
+        let httpServer = Kitura.addHTTPServer(onPort: httpPort ?? self.httpPort, with: router, allowPortReuse: true)
```
djones6: Same question as above - why do we need port reuse?

PR #1310, opened 2018-07-25
Codable parameter inconsistency changes

saiHemak:
## Description
Codable routing inconsistent with parameters

## Motivation and Context
Currently in Kitura you could register the following route:

```
router.delete("/users/:id", handler: deleteAllHandler)

func deleteAllHandler(completion: (RequestError?) -> Void) {
        //Some logic
        completion(nil)
}
```
And Kitura is happy with this.
However if you do the following:
```
router.delete("/users/:id", handler: deleteSingleHandler)

func deleteSingleHandler(id: String, completion: (RequestError?) -> Void) {
        //Some logic
        completion(nil)
    }
```
An error is logged:

[ERROR] [CodableRouter.swift:426 parameterIsPresent(in:)] Erroneous path '/users/:id', parameter ':id' is not allowed. Codable routes do not allow parameters.
This is inconsistent, we allow (but ignore) the /:id parameter in the first case but error on the second case.

There's also a few issues with this:

In the first case, if someone makes a request to users/3 it could allow this request to delete all users.

When using any of the Codable routes with an identifier, the code checks to see if the endpoint has any parameters, for example /:some-Id and then errors even if the parameter provided is actually /:id. Which is actually the correct parameter, and is the one Kitura appends for you.
## How Has This Been Tested?
Ran Kitura test cases

## Checklist:


saiHemak:
@ddunn2  can you please review

Sources/Kitura/CodableRouter.swift
```diff
@@ -539,6 +539,9 @@ extension Router {
 
     // DELETE
     fileprivate func deleteSafely(_ route: String, handler: @escaping NonCodableClosure) {
+        if !isValidParameter(of: route, isTolerable: false) {
```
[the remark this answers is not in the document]
saiHemak: how about isAllowed ?

Sources/Kitura/CodableRouter.swift
```diff
@@ -562,6 +565,25 @@ extension Router {
...
 
+    internal func isValidParameter(of route: String, isTolerable: Bool) -> Bool {
+        if route.contains(":") {
+            if !isTolerable {
+                Log.error("Erroneous path '\(route)', parameter is not allowed. Codable routes without identifier do not allow parameters.")
+                return false
+            } else {
+                let paramaterString = route.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
+                let parameter = paramaterString.count > 0 ? paramaterString[1] : ""
+                if (parameter.isEmpty || parameter.elementsEqual("id")) {
```
djones6: This seems like a really long-winded way to express `route.hasSuffix(":id")` ... but I suppose it _does_ also check that the route doesn't also contain arbitrary path parameters.

I think the whole thing could be clearer in its intent expressed (something like) this:
```swift
/// Precondition: Path is known to contain a : character
func pathHasSingleParamIdAsSuffix(_ path: String) -> Bool {
  let paramaterString = route.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
  let parameter = paramaterString.count > 0 ? paramaterString[1] : ""
  return parameter.isEmpty || parameter.elementsEqual("id")
}

func pathSyntaxIsValid(_ path: String, identifierExpected: Bool) -> Bool {
  let parameterSupplied = path.contains(":")
  let validIdentifierSyntaxUsed = parameterSupplied && pathHasSingleParamIdAsSuffix(path)
  if parameterSupplied && !validIdentifierSyntaxUsed {
    // Log error: arbitrary params not allowed
    return false
  }
  if identifierExpected && !parameterSupplied {
    // Warn in this case?
    return true
  }
  if !identifierExpected && identifierSupplied {
    // Log error: id supplied but route doesn't consume one
    return false
  }
  return true
}
```
note, I didn't test this :smile:

Sources/Kitura/CodableRouter.swift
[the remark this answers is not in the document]
djones6: My thoughts on naming:
1. `isValidParameter(of:)` - The naming doesn't sit well with me. I read the code and ask myself "which parameter are we talking about here?". What this is telling you is whether it is valid for the route string to contain parameters - in fact, one specific parameter: `:id` which must appear at the end of the line.  So how about `pathSyntaxIsValid`?
2. `isTolerable` - The purpose of this parameter is to indicate whether the route handler expects an Identifier, ie. whether we'd expect to find (or add) an `:id` suffix.  So why not name it what it is?  `identifierExpected`

With the two changes above, you'd have something like:
```
if !pathSyntaxIsValid(route, identifierExpected: false) {
    return
}
```

Sources/Kitura/CodableRouter.swift
saiHemak: Thanks @djones6  ..Will address your comments

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,21 +603,38 @@ extension Router {
...
-            Log.error("Erroneous path '\(route)', parameter ':\(parameter)' is not allowed. Codable routes do not allow parameters.")
+    /// Precondition: Path is known to contain a : character
+    func pathHasSingleParamIdAsSuffix(_ path: String) -> Bool {
+        let paramaterString = path.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
+        let parameter = paramaterString.count > 0 ? paramaterString[1] : ""
+        return parameter.isEmpty || parameter.elementsEqual("id")
+    }
+
+    func pathSyntaxIsValid(_ path: String, identifierExpected: Bool) -> Bool {
+        let identifierSupplied = path.contains(":")
```
saiHemak: @djones6 I have re-ordered the checks based on the ```identifierExpected``` flag as  it is vain to do the split operation if the identifier itself is not expected for the route

saiHemak:
@ddunn2 and @djones6  Thanks for your review. I have addressed your comments .. Please review

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,21 +603,38 @@ extension Router {
...
+        if !identifierExpected && identifierSupplied {
+            Log.error("Erroneous path '\(path)' is not allowed. Codable routes do not allow parameters.")
+            return false
+        }
+        if identifierExpected && !identifierSupplied {
+            Log.warning("Identifier expected for the '\(path)'. Id parameter will be automatically appeneded by default to the path .")
```
djones6: `Identifier expected for '\(path)'.` (remove 'the')
`appeneded` -> `appended`
We don't really need "`by default`".  The last statement could simply be: `:id will be automatically appended to the path.`

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,21 +603,38 @@ extension Router {
...
             return true
         }
-        return false
+        let validIdentifierSyntaxUsed = identifierSupplied && pathHasSingleParamIdAsSuffix(path)
+        if identifierExpected && !validIdentifierSyntaxUsed {
+            Log.error("Erroneous path '\(path)'  is not allowed. Codable routes allow only id as parameter.")
```
djones6: double space after `'\(path)'`
Also I'd suggest: `Codable routes allow only id as parameter.` -> `Codable routes support a trailing id parameter only.`

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,21 +603,38 @@ extension Router {
...
+            return false
+        }
+
```
djones6: unnecessary new line

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,21 +603,38 @@ extension Router {
...
+    func pathHasSingleParamIdAsSuffix(_ path: String) -> Bool {
+        let paramaterString = path.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
+        let parameter = paramaterString.count > 0 ? paramaterString[1] : ""
+        return parameter.isEmpty || parameter.elementsEqual("id")
+    }
+
+    func pathSyntaxIsValid(_ path: String, identifierExpected: Bool) -> Bool {
+        let identifierSupplied = path.contains(":")
+        if !identifierExpected && identifierSupplied {
+            Log.error("Erroneous path '\(path)' is not allowed. Codable routes do not allow parameters.")
```
djones6: The path is not necessarily erroneous, just that we don't support arbitrary params (yet).  I might word it as: `Path '\(path)' is not allowed: Codable routes do not allow path parameters.`

saiHemak:
@djones6  Thanks for the review.Modified error messages accordingly.

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,21 +603,37 @@ extension Router {
         }
     }
 
-    internal func parameterIsPresent(in route: String) -> Bool {
-        if route.contains(":") {
-            let paramaterString = route.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
-            let parameter = paramaterString.count > 0 ? paramaterString[1] : ""
-            Log.error("Erroneous path '\(route)', parameter ':\(parameter)' is not allowed. Codable routes do not allow parameters.")
+    /// Precondition: Path is known to contain a : character
+    func pathHasSingleParamIdAsSuffix(_ path: String) -> Bool {
```
ianpartridge: As this function is only used from `pathSyntaxIsValid()` can we make it a [nested function](https://docs.swift.org/swift-book/LanguageGuide/Functions.html#ID178)?

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,21 +603,37 @@ extension Router {
...
+        let paramaterString = path.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
```
ianpartridge: paramaterString -> parameterString

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,21 +603,37 @@ extension Router {
...
+        let parameter = paramaterString.count > 0 ? paramaterString[1] : ""
+        return parameter.isEmpty || parameter.elementsEqual("id")
+    }
+
+    func pathSyntaxIsValid(_ path: String, identifierExpected: Bool) -> Bool {
+        let identifierSupplied = path.contains(":")
+        if !identifierExpected && identifierSupplied {
```
ianpartridge: Would this be clearer if we did

```swift
switch (identifierExpected, identifierSupplied) {
    ...
}
```

and pattern-matched?

saiHemak: @ianpartridge   The above code would look like this if we change it to switch-case 
```
switch(identifierExpected, identifierSupplied) {
        case(false, true) :
            Log.error("Path '\(path)' is not allowed: Codable routes do not allow path parameters.")
            return false
        case (true, false):
            Log.warning("Identifier expected for '\(path)'. :id will be automatically appended to the path.")
            return true
        case (true, true):
            guard pathHasSingleParamIdAsSuffix(path) else {
            Log.error("Erroneous path '\(path)' is not allowed. Codable routes support a trailing id parameter only.")
            return false
            }
            return true
        case (false, false):
            return true
        }
```
However, for two reasons I was thinking not to go for it is , there is no logical reason behind  coupling```identifierExpected``` and  ```identifierSupplied ``` as a tuple. Also, I believe  the readability is not better than the legacy if-else approach. Please let me know which approach we need to take ..
@djones6 please let us know your opinion too ..

Sources/Kitura/CodableRouter.swift
saiHemak: I agree ..however, I would feel that it may readability.  How about marking it private and restricting its access ?

Sources/Kitura/CodableRouter.swift
ianpartridge: Agreed, please make this change. Put the common `(false, false)` fast-path as the first case.

Sources/Kitura/CodableRouter.swift
ianpartridge: I don't see how it would affect readability. This is what nested functions are for - small helpers used by another function.

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,14 +603,31 @@ extension Router {
...
 
-    internal func parameterIsPresent(in route: String) -> Bool {
-        if route.contains(":") {
-            let paramaterString = route.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
-            let parameter = paramaterString.count > 0 ? paramaterString[1] : ""
-            Log.error("Erroneous path '\(route)', parameter ':\(parameter)' is not allowed. Codable routes do not allow parameters.")
+    func pathSyntaxIsValid(_ path: String, identifierExpected: Bool) -> Bool {
+        let identifierSupplied = path.contains(":")
+        switch(identifierExpected, identifierSupplied) {
+        case(false, true) :
```
ianpartridge: Space after `case` please.

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,14 +603,31 @@ extension Router {
...
     }
 
-    internal func parameterIsPresent(in route: String) -> Bool {
-        if route.contains(":") {
-            let paramaterString = route.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
-            let parameter = paramaterString.count > 0 ? paramaterString[1] : ""
-            Log.error("Erroneous path '\(route)', parameter ':\(parameter)' is not allowed. Codable routes do not allow parameters.")
+    func pathSyntaxIsValid(_ path: String, identifierExpected: Bool) -> Bool {
+        let identifierSupplied = path.contains(":")
+        switch(identifierExpected, identifierSupplied) {
```
ianpartridge: Space after `switch` please.

Sources/Kitura/CodableRouter.swift
ianpartridge: No space before `:`.

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,14 +603,31 @@ extension Router {
...
+            return false
+        case (true, false):
+            Log.warning("Identifier expected for '\(path)'. :id will be automatically appended to the path.")
+            return true
+        case (true, true):
+            /// Precondition: Path is known to contain a : character
+            func pathHasSingleParamIdAsSuffix(_ path: String) -> Bool {
+                let parameterString = path.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
+                let parameter = parameterString.count > 0 ? parameterString[1] : ""
+                return parameter.isEmpty || parameter.elementsEqual("id")
```
ianpartridge: Why `elementsEqual`? Can't we just use `==`?

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,14 +603,31 @@ extension Router {
...
+            }
+
+            guard pathHasSingleParamIdAsSuffix(path) else {
+            Log.error("Erroneous path '\(path)' is not allowed. Codable routes support a trailing id parameter only.")
```
ianpartridge: Indentation.

ianpartridge:
Thanks @saiHemak. Please just push another commit to your branch, don't force-push. It's easier to review if each review cycle is a new commit. We squash when we merge.

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -886,4 +888,31 @@ class TestCodableRouter: KituraTest {
...
+            .hasStatus(.notFound)
+            .hasData()
+            .run()
+    }
+
+    func testIdentifierNotExpected() {
+        //Add this erroneous route which should not be hit by the test, should log an error but we can't test the log so we check for a 404 not found.
+        router.delete("/users/:id") { (respondWith: (RequestError?) -> Void) in
+            print("DELETE on /users")
+             respondWith(RequestError(.badRequest, body: Conflict(on: "id")))
```
ianpartridge: Indentation.

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,14 +603,31 @@ extension Router {
...
+            /// Precondition: Path is known to contain a : character
+            func pathHasSingleParamIdAsSuffix(_ path: String) -> Bool {
+                let parameterString = path.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
+                let parameter = parameterString.count > 0 ? parameterString[1] : ""
+                return parameter.isEmpty || parameter == "id"
+            }
+
+            guard pathHasSingleParamIdAsSuffix(path) else {
+                Log.error("Erroneous path '\(path)' is not allowed. Codable routes support a trailing id parameter only.")
+            return false
```
ianpartridge: Indentation.

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,14 +603,31 @@ extension Router {
...
+        case (true, false):
+            Log.warning("Identifier expected for '\(path)'. :id will be automatically appended to the path.")
+            return true
+        case (true, true):
+            /// Precondition: Path is known to contain a : character
+            func pathHasSingleParamIdAsSuffix(_ path: String) -> Bool {
+                let parameterString = path.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
+                let parameter = parameterString.count > 0 ? parameterString[1] : ""
+                return parameter.isEmpty || parameter == "id"
```
ianpartridge: What happens if the path is `/hello/:`, i.e. it ends with a colon? If I am reading the code correctly the function will return true which doesn't seem right? Please add a test for a route that ends with `:`.

saiHemak:
Had detailed discussion with @ianpartridge  and here is the behavior expected behavior when identifier is expected
````/status` -> `/status/:id`
`/status/` -> `/status/:id`
`/status/:` -> error
`/status/:id` -> `/status/:id`
`/status/:foo` -> error
```
As per the above understanding  ``` return parameter.isEmpty || parameter == "id" ```  should return true only if parameter name is ```id```

Sources/Kitura/CodableRouter.swift
```diff
@@ -600,14 +603,31 @@ extension Router {
...
+            Log.error("Path '\(path)' is not allowed: Codable routes do not allow path parameters.")
+            return false
+        case (true, false):
+            Log.warning("Identifier expected for '\(path)'. :id will be automatically appended to the path.")
+            return true
+        case (true, true):
+            /// Precondition: Path is known to contain a : character
+            func pathHasSingleParamIdAsSuffix(_ path: String) -> Bool {
+                let parameterString = path.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
+                let parameter = parameterString.count > 0 ? parameterString[1] : ""
```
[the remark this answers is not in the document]
ianpartridge: I'm talking about this specific function, not the broader code.  You do `path.split()` with `maxSplits: 1`. That means theoretically it could do 0 splits. In which case `parameterString.count` would be 1 and we would crash.

Also, I don't understand why we pass `omittingEmptySubsequences: false`. Why would we want empty subsequences to be created?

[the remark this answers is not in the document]
saiHemak: Yeah it would crash if you call this function on a String which does not have the ```:```  and hence  the comment ```Precondition: Path is known to contain a : character ``` has  been retained. But in  this case it will not happen as we would not end up in calling this function if the path does not have  a ```:``` character and the possibility of ending up in performing ```0``` splits is not there . Also, as this  method's scope is  limited to this ```case``` one will not end up in calling it without the ```contains(":")``` check from outside code. Moreover, ```parameterIsPresent``` in the  existing  code is also doing the same.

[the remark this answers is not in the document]
ianpartridge: Right, as I noted, it won't crash today because of the circumstances we call the function under. But who knows about the future? We should practice [Defensive Programming](https://en.wikipedia.org/wiki/Defensive_programming) and not store up problems for the future.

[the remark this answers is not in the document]
saiHemak: If  ```omittingEmptySubsequences```  is  set to false an empty subsequence is returned in the result for each consecutive pair of ```:``` . Its useful in cases of handling the case like ```/status/:``` where empty sub-sequence would be returned and will not end up in crash while doing comparison. 
eg :  In case of  ```omittingEmptySubsequences: false```
```
let path = "/status/:"
let parameterString = path.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
print(parameterString.count)  ------------>  would be 2 
let parameter = parameterString.count > 0 ? parameterString[1] : ""
print(parameter) -----------> would be ""
``` 
In case of ```omittingEmptySubsequences: true``` , an empty sub-sequence would not be returned.
```
let path = "/status/:"
let parameterString = path.split(separator: ":", maxSplits: 1, omittingEmptySubsequences: false)
print(parameterString.count) ------------> would be 1
let parameter = parameterString.count > 0 ? parameterString[1] : "" // -----------> CRASH 
print(parameter)
```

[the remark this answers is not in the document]
saiHemak: In that case ``` if route.contains(":") {``` redundant check needs to be added in the method  ..Please let me know if you want me to do that !?

djones6:
@ddunn2 Are you happy with the changes for this PR now? Could you re-review?

Sources/Kitura/CodableRouter.swift
```diff
@@ -604,14 +607,33 @@ extension Router {
...
+        switch (identifierExpected, identifierSupplied) {
+        case (false, true):
+            Log.error("Path '\(path)' is not allowed: Codable routes do not allow path parameters.")
+            return false
+        case (true, false):
+            Log.warning("Identifier expected for '\(path)'. :id will be automatically appended to the path.")
+            return true
+        case (true, true):
+            /// Precondition: Path is known to contain a : character
+            func pathHasSingleParamIdAsSuffix(_ path: String) -> Bool {
```
djones6: It seems _weird_ to be declaring a function within a case statement... if this function has no value outside the block, couldn't the logic just be inlined?

Sources/Kitura/CodableRouter.swift
```diff
@@ -430,18 +433,23 @@ extension Router {
 
     // Get array
     fileprivate func getSafely<O: Codable>(_ route: String, handler: @escaping CodableArrayClosure<O>) {
-        registerGetRoute(route: route, outputType: O.self, outputIsArray: true)
+        if !pathSyntaxIsValid(route, identifierExpected: false) {
+            return
+        }
+        registerGetRoute(route: route, outputType: O.self)
```
djones6: You've lost the `outputIsArray: true` (which is why a test is failing)

Sources/Kitura/CodableRouter.swift
```diff
@@ -430,18 +433,23 @@ extension Router {
...
-    fileprivate func getSafely<Id: Identifier, O: Codable>(_ route: String, handler: @escaping IdentifierCodableArrayClosure<Id, O>) {
-        // FIXME: The ID is returned in a tuple, it is not an input parameter
-        // https://github.com/IBM-Swift/Kitura/issues/1336
-        registerGetRoute(route: route, id: Id.self, outputType: O.self, outputIsArray: true)
+    fileprivate func getSafely<Id: Identifier, O: Codable>(_ route: String, handler: @escaping
+        IdentifierCodableArrayClosure<Id, O>) {
+        if !pathSyntaxIsValid(route, identifierExpected: false) {
+            return
+        }
+        registerGetRoute(route: route, id: Id.self, outputType: O.self)
```
djones6: Same here

Issue #1307, opened 2018-07-09
RouterResponse.send(String?) unexpected behaviour


djones6:
## Context and Description
When using 'raw' routing, `RouterResponse` provides two unlabelled `send()` functions:
1. [`send(_ str: String)`](https://github.com/IBM-Swift/Kitura/blob/2.4.1/Sources/Kitura/RouterResponse.swift#L210) that writes raw text to the response body
2. [`send<T : Encodable>(_ obj: T)`](https://github.com/IBM-Swift/Kitura/blob/2.4.1/Sources/Kitura/RouterResponse.swift#L497) that attempts to JSON-encode an Encodable payload.

The expected usage is for a user to send a `String` to send plain text responses, or a custom type that conforms to `Codable` when sending JSON responses.

If one attempts to send a `String?` instead, this does not match the first signature, and instead matches the `Encodable` signature.  Because a String on its own does not represent a valid JSON document, the JSON encoding will fail, resulting in an empty response body and a warning logged.

We should permit sending of raw text when a `String?` is passed, unwrapping the optional value and writing the string data as before.

### Considerations for handling `nil`
We need to define the behavior for when a user passes a `nil` value to `send(_ str: String?)`.  Some possibilities:
1. Accept it as a valid representation of an empty body, and leave the body unchanged.
2. As **1**, but log a warning message that `send` was called with a `nil` value.
3. Literally write `nil` to the body.

I'd suggest that **2** is preferable, but all have their pro's and con's. I think it's more likely a user expects to be sending data rather than nil, and to do so is a mistake (and so they would want to know about it).  To avoid logging the warning, the user could unwrap the value themselves before invoking send, and simply skip the send on nil.

djones6:
This was resolved by #1313 (Kitura 2.5.0)

PR #1313, opened 2018-07-26
Add optional String function to allow passing nil Strings

ShihabMehboob:
Added an extra routerResponse function that takes in optional Strings to allow nil Strings to be sent without resorting to matching the second unlabelled JSON object function. In relation to this issue: https://github.com/IBM-Swift/Kitura/issues/1307


Sources/Kitura/RouterResponse.swift
```diff
@@ -222,6 +222,28 @@ public class RouterResponse {
...
+    ///
+    /// - Parameter str: the string to send.
+    /// - Returns: this RouterResponse.
+    @discardableResult
+    public func send(_ str: String?) -> RouterResponse {
+        guard !state.invokedEnd else {
+            Log.warning("RouterResponse send(str:) invoked after end() for \(self.request.urlURL)")
+            return self
+        }
+        var stringToSend: String
```
ianpartridge: Couldn't this whole thing just be:

```swift
guard let str = str else {
    Log.warning("RouterResponse send(str:) invoked with nil")
    return send("")
}
return send(str)
```

.swift-version
```diff
@@ -1 +1 @@
-4.1.2
+4.1
```
ianpartridge: ?

Tests/KituraTests/TestResponse.swift
```diff
@@ -1493,6 +1493,8 @@ class TestResponse: KituraTest {
 
                 try response.send(status: HTTPStatusCode.forbidden).send(data: "<!DOCTYPE html><html><body><b>forbidden</b></body></html>\n\n".data(using: .utf8)!).end()
                 try response.send(status: HTTPStatusCode.OK).end()
+                let emptyString: String? = nil
```
ianpartridge: This should be named `nilString`.

Sources/Kitura/RouterResponse.swift
```diff
@@ -222,6 +222,19 @@ public class RouterResponse {
...
     }
+    
+    /// Send an optional string.
+    ///
+    /// - Parameter str: the string to send.
+    /// - Returns: this RouterResponse.
+    @discardableResult
+    public func send(_ str: String?) -> RouterResponse {
+        guard !state.invokedEnd, let str = str else {
+            Log.warning("RouterResponse send(str:) invoked after end() for \(self.request.urlURL)")
```
djones6: This doesn't look right, you've rolled two different checks into one guard, when they are checking two distinct cases:
- `state.invokedEnd` should log a warning ("send invoked after end") and return `self`,
- `str == nil` should log a different warning ("send invoked with a nil string") and return `send("")`.

ShihabMehboob:
@ianpartridge Thanks, made those test changes now to be more clear.

Tests/KituraTests/TestResponse.swift
```diff
@@ -101,6 +103,38 @@ class TestResponse: KituraTest {
...
     }
+    
+    func testOptionalStringNilResponse() {
+        performServerTest(router) { expectation in
+            self.performRequest("post", path:"/sendNilString", callback: { response in
+                XCTAssertEqual(response?.statusCode, HTTPStatusCode.OK, "HTTP Status code was \(String(describing: response?.statusCode))")
+                
+                do {
+                    let body = try response?.readString()
+                    XCTAssertNil(body)
```
ianpartridge: I thought in the nil case we were sending the empty string - so wouldn't we expect to get an empty body back?

ShihabMehboob:
@ianpartridge Realised that the default readString() function would always return nil if the String was empty, opened a PR in Kitura-Net to amend this: https://github.com/IBM-Swift/Kitura-net/pull/274

ShihabMehboob:
@ianpartridge Sure, amended the test to use readAllData() instead, removing the reliance on the new Kitura-net API.

PR #1314, opened 2018-08-03
Content type

Andrew-Lees11:
## Description
Add a wrapper around the MediaType (e.g "application/json") from the content-type header so that when we use it in custom encoders/decoders it is not a string.

## How Has This Been Tested?
Tests have been added covering expected behaviour and current tests still pass.

## Checklist:

### DIFF
--- a/Sources/Kitura/contentType/MediaType.swift
+++ b/Sources/Kitura/contentType/MediaType.swift
@@ -16,30 +16,55 @@
 
 import KituraNet
 
+/**
+ The media type (formerly known as MIME type) is a standardized way to indicate the nature and format of a document.
+ This struct consists of a catagorical `topLevelType` and specific `subtype` seperated by a "/" (e.g. "text/plain").
+ In HTTP, The media type is sent as the first section of the "Content-Type" header and is case insensitive.
+ ### Usage Example: ###
+ ```swift
+ let mediaType = MediaType(type: .application, subtype: "json")
+ print(mediaType.description)
+ // Prints ("application/json")
+ ```
+ */
 public struct MediaType: CustomStringConvertible {
     
+    /// An enum of all media type's top level types.
     public enum TopLevelType: String {
-        case application = "application"
-        case audio = "audio"
-        case font = "font"
-        case image = "image"
-        case message = "message"
-        case model = "model"
-        case multipart = "multipart"
-        case text = "text"
-        case video = "video"
+        case application, audio, font, image, message, model, multipart, text, video
     }
     
-    let topLevelType: TopLevelType
+    // MARK: Media type components
     
-    let subtype: String
+    /// The topLevelType represents the category of the MediaType (e.g. "audio").
+    public let topLevelType: TopLevelType
     
-    init(type: TopLevelType, subtype: String = "*") {
+    /// The subtype is the specific MediaType (e.g. "html").
+    public let subtype: String
+    
+    // MARK: Initializers
+    
+    /**
+     Initialize a `MediaType` instance with a `TopLevelType` type and `String` subtype. If no subtype is provided it will default to "*" representing all subtypes.
+     ### Usage Example: ###
+     ```swift
+     let mediaType = MediaType(type: .application, subtype: "json")
+     ```
+     */
+    public init(type: TopLevelType, subtype: String = "*") {
         self.topLevelType = type
         self.subtype = subtype.lowercased()
     }
     
-    init? (_ mimeType: String) {
+    /**
+     Initialize a `MediaType` instance from a `String`. If no subtype is provided it will default to "*" representing all subtypes.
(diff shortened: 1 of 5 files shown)


Sources/Kitura/contentType/MediaType.swift
```diff
@@ -0,0 +1,147 @@
...
+ This struct consists of a catagorical `topLevelType` and specific `subtype` seperated by a "/" (e.g. "text/plain").
+ In HTTP, The media type is sent as the first section of the "Content-Type" header and is case insensitive.
+ ### Usage Example: ###
+ ```swift
+ let mediaType = MediaType(type: .application, subtype: "json")
+ print(mediaType.description)
+ // Prints ("application/json")
+ ```
+ */
+public struct MediaType: CustomStringConvertible, Equatable, Hashable {
```
ianpartridge: Why is this `public`? Seems like something we're going to use internally inside Kitura but not expose to users.

Sources/Kitura/contentType/MediaType.swift
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     Initialize a `MediaType` instance from a `KituraNet` `HeadersContainer`. This will extract the media type String from the first section of the "Content-type" header and use this to initialize itself. If the string preceding the first "/" is not a valid `TopLevelType` the initializer will return nil.
+     ### Usage Example: ###
+     ```swift
+     let headers = HeadersContainer()
+     headers.append("Content-Type", value: ["application/json"])
+     let mediaType = MediaType(headers: headers)
+     ```
+     */
+    public init? (headers: HeadersContainer) {
```
ianpartridge: We shouldn't expose KituraNet API out of Kitura, so this should be `internal`.

Sources/Kitura/contentType/MediaType.swift
```diff
@@ -0,0 +1,147 @@
...
+     // Prints ("application/json")
+     ```
+     */
+    public var description: String {
+        return "\(topLevelType)/\(subtype)"
+    }
+    
+    /// Compares two MediaTypes returning true if they are equal. Required for Equatable conformance.
+    public static func == (lhs: MediaType, rhs: MediaType) -> Bool {
+        return lhs.description == rhs.description
```
ianpartridge: Probably faster to do:

```swift
guard lhs.topLevelType == rhs.topLevelType else {
    return false
}
return lhs.subtype == rhs.subtype
```

Sources/Kitura/contentType/MediaType.swift
```diff
@@ -0,0 +1,147 @@
...
+    /**
+     Returns the media type, as a String structured: `topLevelType`/`subtype`. Required for CustomStringConvertible conformance.
+     ### Usage Example: ###
+     ```swift
+     print(mediaType.description)
+     // Prints ("application/json")
+     ```
+     */
+    public var description: String {
+        return "\(topLevelType)/\(subtype)"
```
ianpartridge: This will recalculate the description every time the property is accessed. It would be better to use `didSet{}` on topLevelType and subtype to recalculate the String only when necessary.

Sources/Kitura/contentType/MediaType.swift
```diff
@@ -0,0 +1,147 @@
...
+        case application, audio, font, image, message, model, multipart, text, video
+    }
+    
+    // MARK: Media type components
+    
+    /// The topLevelType represents the category of the MediaType (e.g. "audio").
+    public let topLevelType: TopLevelType
+    
+    /// The subtype is the specific MediaType (e.g. "html").
+    public let subtype: String
```
ianpartridge: Can this be `subType` please?

Sources/Kitura/contentType/MediaType.swift
```diff
@@ -0,0 +1,147 @@
...
+    // MARK: Initializers
+    
+    /**
+     Initialize a `MediaType` instance with a `TopLevelType` type and `String` subtype. If no subtype is provided it will default to "*" representing all subtypes.
+     ### Usage Example: ###
+     ```swift
+     let mediaType = MediaType(type: .application, subtype: "json")
+     ```
+     */
+    public init(type: TopLevelType, subtype: String = "*") {
```
ianpartridge: subtype -> subType

Sources/Kitura/contentType/MediaType.swift
```diff
@@ -0,0 +1,147 @@
...
+    }
+    
+    /// Compares two MediaTypes returning true if they are equal. Required for Equatable conformance.
+    public static func == (lhs: MediaType, rhs: MediaType) -> Bool {
+        return lhs.description == rhs.description
+    }
+    
+    /// The hashValue for the MediaTypes. Required for Hashable conformance.
+    public var hashValue: Int {
+        return description.hashValue
```
ianpartridge: This is another reason to cache `description`. Currently accessing the hash value will recalculate the description.

PR #1316, opened 2018-08-07
Improve type handling for id params in generated swagger

djones6:
## Description

Reworking of the handling of the id parameter, to permit both Int and String types rather than hard-coding `integer`.

This is code from @nhardman in #1304 which I have split into smaller PRs.

## How Has This Been Tested?

This adds new Swagger generation tests for both String and Int type identifiers.

## Checklist:

### DIFF
--- a/Sources/Kitura/SwaggerGenerator.swift
+++ b/Sources/Kitura/SwaggerGenerator.swift
@@ -316,11 +316,21 @@ struct SwaggerDocument: Encodable {
     // - Returns: SwaggerParameter.
     func buildParameter(name: String, parametertype: String) -> SwaggerParameter {
         if name == "id" {
+            // id can only be string or integer. so force the type to string if
+            // it is neither.
+            var idType = "string"
+            if let ptype = SwiftType(rawValue: parametertype) {
+                idType = ptype.swaggerType()
+                if idType != "integer" && idType != "string" {
+                    idType = "string"
+                }
+            }
+
             return SwaggerParameter(invalue: "path",
                                     name: name,
                                     required: true,
                                     schema: nil,
-                                    type: "integer")
+                                    type: idType)
         }
         return SwaggerParameter(invalue: "body",
                                 name: name,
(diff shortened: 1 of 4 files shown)


Tests/KituraTests/TestSwaggerGeneration.swift
```diff
@@ -628,6 +633,69 @@ class TestSwaggerGeneration: KituraTest {
         }
     }
 
+    func pathContentAssertions4(json: String?) {
+        if let jsonString = json {
```
ianpartridge: `guard let` does exist ;)

Sources/Kitura/SwaggerGenerator.swift
```diff
@@ -316,11 +316,21 @@ struct SwaggerDocument: Encodable {
     // - Returns: SwaggerParameter.
     func buildParameter(name: String, parametertype: String) -> SwaggerParameter {
         if name == "id" {
+            // id can only be string or integer. so force the type to string if
+            // it is neither.
+            var idType = "string"
+            if let ptype = SwiftType(rawValue: parametertype) {
```
ianpartridge: I think I understand what's going on here, but it would be nice to make a few improvements:

1. `parametertype` -> `parameterType`
2. `swaggerType()` should be a property not a function
3. If the name can only be String or Integer why not define that in an enum rather than passing Strings everywhere

djones6:
@ianpartridge I've addressed the review comments, except for replacing passing of strings with an enum: this bit of code needs re-working back up to the Router level to pass types through rather than string interpolating the type names.  I'd rather do that in a separate PR.

djones6:
Travis is having issues (https://t.co/IX0wg2yH0J) so our Linux builds seem to be failing frequently.

djones6:
@ianpartridge I've backed out the review changes so that this + the follow-on PRs can merge cleanly.  I'll implement your suggestions in the final one.

PR #1324, opened 2018-08-19
Allow users to switch to using the NIO port of KituraNet

pushkarnk:
We recently renamed the erstwhile KituraNIO package to KituraNet. So, we now have a `KituraNet` and a `SwiftNIO-based KituraNet`. This pull request allows Kitura users to switch to using the latter using only an environment variable named `KITURA_NIO`.

## Description
The core of the change is only to the Swift package manifest file. In the building phase of the Kitura application, if an environment variable named `KITURA_NIO` is set, the Swift package manager switches to using the NIO-based KituraNet found [here](https://github.com/IBM-Swift/KituraNIO). The other changes in this pull request are in the Tests and mostly around two limitations of the NIO-based port - [a bind issue ](https://github.com/IBM-Swift/KituraNIO/issues/2)and the absence of a FastCGI implementation. 

To enable the use of the `SwiftNIO-based KituraNet` we use:
`export KITURA_NIO=1`

To switch back to using stock KituraNet in the same environment:
`unset KITURA_NIO`

## Motivation and Context
Support for a NIO-based networking package was available until now only on a branch named `kitura-nio`. This pull request makes it available on the master branch.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,6 +33,11 @@ matrix:
     - os: linux
       dist: trusty
       sudo: required
+      services: docker
+      env: DOCKER_IMAGE=ubuntu:16.04 KITURA_NIO=1
+    - os: linux
+      dist: trusty
+      sudo: required
       env: SWIFT_SNAPSHOT=swift-4.2-DEVELOPMENT-SNAPSHOT-2018-06-19-a
     - os: linux
       dist: trusty
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,10 @@ matrix:
       dist: trusty
       sudo: required
       env: SWIFT_SNAPSHOT=swift-4.2-DEVELOPMENT-SNAPSHOT-2018-06-19-a
+    - os: linux
+      dist: trusty
+      sudo: required
+      env: SWIFT_SNAPSHOT=swift-4.2-DEVELOPMENT-SNAPSHOT-2018-06-19-a KITURA_NIO=1
     - os: osx
       osx_image: xcode9.2
       sudo: required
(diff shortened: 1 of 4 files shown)


pushkarnk:
I added CI runs with KITURA_NIO=1 on trusty, xenial(docker) and osx to the travis.yml. And we've hit the [linker error](https://github.com/IBM-Swift/Kitura-NIO/issues/62).

pushkarnk:
I've removed CI builds on Ubuntu 16.04 while we try to address the [linker error](url
https://github.com/IBM-Swift/Kitura-NIO/issues/62) which has started to look like an Ubuntu thing.

.travis.yml
```diff
@@ -46,6 +50,13 @@ matrix:
       osx_image: xcode10
       sudo: required
       env: SWIFT_SNAPSHOT=4.2
+    - os: osx
+      osx_image: xcode10
+      sudo: required
+      env: SWIFT_SNAPSHOT=4.2 KITURA_NIO=1
+
+before_install:
+  - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then brew update && brew install libressl ; fi
```
ianpartridge: I guess this is going to be required for all Kitura packages on macOS? If so, perhaps it would be better added to Package-Builder.

Package.swift
```diff
@@ -18,6 +18,15 @@
  **/
 
 import PackageDescription
+import Foundation
+
+var kituraNetPackage: Package.Dependency
+
+if let _ = ProcessInfo.processInfo.environment["KITURA_NIO"] {
+    kituraNetPackage = .package(url: "https://github.com/IBM-Swift/Kitura-NIO.git", from: "1.0.0")
```
ianpartridge: I wonder if there is a case for tagging Kitura-NIO to 2.0.0, just to keep things closer in sync?

Package.swift
```diff
@@ -18,6 +18,15 @@
  **/
 
 import PackageDescription
+import Foundation
+
+var kituraNetPackage: Package.Dependency
+
+if let _ = ProcessInfo.processInfo.environment["KITURA_NIO"] {
```
ianpartridge: I think I prefer `if ProcessInfo.processInfo.environment["KITURA_NIO"] != nil` like you did in the tests.

Package.swift
pushkarnk: any comments on that @djones6 ?

.travis.yml
pushkarnk: Sure, will do

Issue #1332, opened 2018-09-11
StaticFileServer not sending 304 Not Modified when local file has matching Etag


CalebKussmaul:
## Context and Description
When a browser sends a conditional GET to a server, the server may respond with 304 NOT MODIFIED and the version of the resource in the browser's cache will be used instead. This allows a requested resource to only be sent when the version that the browser already has is out of date or does not exist. This is accomplished by the browser including the cached file's modification date, or a value associated with the file's contents (Etag) in its request to the server.

Additionally, a cache control value for max-age can be sent to the browser with a resource, and the browser will not make any request to the server unless it detects the resource may be expired. In this case it will still send a conditional GET request and rely on the cached version if it isn't out of date.  

[a good explanation can be found here](https://stackoverflow.com/a/6215091/3236716)

## problem 1: max-age improperly documented
The value for maxAgeCacheControlHeader in the options of the StaticFileServer, is documented as:

`Parameter maxAgeCacheControlHeader: a max-age in milliseconds for "max-age" value in "Cache-Control" header in the response`

The value actually represents seconds, not milliseconds. I have tested it and the documentation is not correct.

## problem 2: Kitura responds with the entire file when the browser's version is not out of date

The following are relevant parts of an exchange using StaticFileServer middleware recorded from Wireshark. I have addLastModifiedHeader set to false so that it should solely rely on the Etag (note that from my testing this did not have any effect on the result), and with maxAgeCacheControlHeader set to 0 so it should always make a conditional GET.

### Expected vs. Actual Behavior
**Expected**:

Request HTTP header:

```
GET /assets/Style.css HTTP/1.1\r\n
...
If-None-Match: W/"1d-5b980193"\r\n
```

Response (assuming Etag matches file):

```
HTTP/1.1 304 NOT MODIFIED\r\n
```
**Actual**:

Request HTTP header:

```
GET /assets/Style.css HTTP/1.1\r\n
...
If-None-Match: W/"1d-5b980193"\r\n
```

Response:

```
HTTP/1.1 200 OK\r\n
...
Etag: W/"1d-5b980193"\r\n
...
<entire content of Style.css>
```

Depending on the service, this could have a major impact on performance and network usage.

PR #1333, opened 2018-09-11
Fix: Implemented Conditional GET using etags

CalebKussmaul:
On non-range requests, kitura StaticFileServer will never send HTTP 304 NOT MODIFIED response, even if client already has the latest version of a resource. It will send the entire file instead, which is a waste of resources. Additionally, a documentation error regarding the max-age header parameter was fixed. 

## Description

A check was added to see if the file's Etag matched the request's corresponding "If-None-Match" parameter. If they are equal, the server will respond with HTTP 304 NOT MODIFIED. Otherwise it will still send the file.

## Motivation and Context

See issue #1332. In some use cases, a browser may request the same resource on many different pages, or multiple times for a number of other reasons. Using the max-age parameter is inadequate because it does not allow the developer to update a resource at any time without clients continuing to use an outdated version of a resource. I created this because I need my server to be as efficient as possible, and under my use case this change results in a significant resource savings. 

## How Has This Been Tested?

I have tested the code on a personal project (on a mac using a browser on the same computer). It should not affect other areas of the code. 

## Checklist:


ianpartridge:
Hi @CalebKussmaul! Are you able to update this PR? We'd like to merge it :)

CalebKussmaul:
@ianpartridge It should be all set now, thank you!

ianpartridge:
@djones6 can you take another look please :)

Sources/Kitura/staticFileServer/FileServer.swift
```diff
@@ -187,9 +187,15 @@ extension StaticFileServer {
                         // Send only headers
                         _ = response.send(status: .OK)
                     } else {
-                        // Send the entire file
-                        try response.send(fileName: filePath)
-                        response.statusCode = .OK
+                        if let etag = request.headers["If-None-Match"],
+                          etag == CacheRelatedHeadersSetter.calculateETag(from: fileAttributes) {
+                            response.statusCode = .notModified
+                            response.send("")
```
djones6: I don't think this empty `send("")` is necessary - it'll do some work under the covers, but result in no data being appended to the response buffer.  Setting the `statusCode` is sufficient.

One of the outcomes of invoking `send()` is that it sets `response.state.invokedSend` to indicate that a response has/is being sent.  However, we only send a default response if `invokedSend` is false _and_ `statusCode` is `.unknown`.

ianpartridge:
Thanks @CalebKussmaul!  @djones6 - please rereview.

PR #1334, opened 2018-09-16
Handle optional query parameters correctly with typesafe middlewares

seabaylea:
## Motivation and Context

Kitura currently doesn't handle optional QueryParameters correctly when used in conjunction with TypedMiddlewares.

If you define a route that has optional query parameters, eg:
```swift
    func getAllHandler(query: ToDoQuery?, completion: @escaping ([ToDo]?, RequestError?) -> Void ) {
        ToDo.findAll(matching: query, completion)
    }
```
the value of `query` is either an instance of a ToDoQuery struct, or `nil`.

However if you add a type-safe middleware, eg for auth:
```swift
    func getAllHandler(auth: MyHTTPAuth, query: ToDoQuery?, completion: @escaping ([ToDo]?, RequestError?) -> Void ) {
        ToDo.findAll(matching: query, completion)
    }
```
then you actually receive an instance of an Optional ToDoQuery, with all the fields set to nil, eg:
```
Optional(Application.ToDoQuery(name: nil, age: nil))
```

As well as not being consistent with the non-typed middleware case, it also doesn't work with the ORM.

This PR fixes this issue for all `GET` requests that return an array (ie, where a QueryParameter might be optional because you might want to get all entries) and for `DELETE` (again, because you might have optional query parameters to either delete a subset, or all entries).

## How Has This Been Tested?
Additional tests have been added for both `GET` and `DELETE` requests for type safe middlewares.

## Checklist:

- [X] If applicable, I have updated the documentation accordingly.

### DIFF
--- a/Sources/Kitura/CodableRouter+TypeSafeMiddleware.swift
+++ b/Sources/Kitura/CodableRouter+TypeSafeMiddleware.swift
@@ -682,6 +682,63 @@ extension Router {
             }
         }
     }
+    
+    /**
+     Sets up a closure that will be invoked when a GET request to the provided route is received by the server.
+     The closure accepts a successfully executed instance of `TypeSafeMiddleware`, the parsed query parameters,
+     and a handler which responds with an array of Codable objects or a `RequestError`.
+     The handler contains the developer's logic, which determines the server's response.
+     ### Usage Example: ###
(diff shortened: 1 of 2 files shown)


seabaylea:
@djones6 If you're happy with this, can you merge?

PR #1338, opened 2018-09-19
Support Swift 4.2

djones6:
Switches to the swift-4.2-RELEASE build (on Linux; macOS is already there as Travis has released an xcode10 GM image with the same name - `xcode10` - as the previous betas).

Also resolves a compilation warning since conditional cast of NSError to Error is no longer required on Linux.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -38,11 +38,11 @@ matrix:
     - os: linux
       dist: trusty
       sudo: required
-      env: SWIFT_SNAPSHOT=swift-4.2-DEVELOPMENT-SNAPSHOT-2018-06-19-a
+      env: SWIFT_SNAPSHOT=4.2
     - os: linux
       dist: trusty
       sudo: required
-      env: SWIFT_SNAPSHOT=swift-4.2-DEVELOPMENT-SNAPSHOT-2018-06-19-a KITURA_NIO=1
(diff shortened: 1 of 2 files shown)


djones6:
Hmm, my `#if !os(Linux) || swift(>=4.2)` is being ignored - the warning is still generated.

It seems that if I have `// swift-tools-version:4.0` at the top of Package.swift, I get the warning (swift is not >=4.2), but if I change that to `// swift-tools-version:4.2`, then the check succeeds and I compile without warnings.

Is this a bug?

djones6:
We need to use `swift(>=4.1.50)` as a method of detecting Swift 4.2 when running in Swift 4.0 compatibility mode (the result of declaring `// swift-tools-version:4.0` in our `Package.swift`).

There is also this: https://github.com/apple/swift-evolution/blob/master/proposals/0212-compiler-version-directive.md - which nicely spells out the combinations of Swift compiler and compatibility modes.  It also implies that, as of Swift 4.2, we could use: `#if compiler(>=4.2)`.

djones6:
OK... we can't use `#compiler` because it's a compile error prior to Swift 4.2 :unamused:

djones6:
Remaining warnings (relating to system packages) need resolving in Kitura-net and BlueSSLService, I'll raise separate PRs for those.

PR #1339, opened 2018-09-25
Improve logging for SwaggerGenerator

djones6:
## Description

Increases the available logging produced by the SwaggerGenerator, to assist in debugging generation failures with complex types.

This uses https://github.com/IBM-Swift/TypeDecoder/pull/12 (now tagged in TypeDecoder 1.2.0)
... and https://github.com/IBM-Swift/TypeDecoder/pull/13 (now tagged in TypeDecoder 1.3.0)

## Motivation and Context

There are several instances where we currently swallow errors while generating swagger, which makes it difficult to debug failures.  This PR logs errors at `error` level, increases the level of processing related messages (generation of models and paths) to `verbose`, and clarifies wording for some of the existing `debug` messages.

One implication of this change is that errors encountered during swagger generation will be logged for all Kitura users, regardless of whether they wanted to generate swagger.  We may need to think about whether this is appropriate, or whether we should have a way to opt in/out of the swagger generation capability.

## Checklist:


Sources/Kitura/SwaggerGenerator.swift
```diff
@@ -449,9 +449,12 @@ struct SwaggerDocument: Encodable {
     func isDictEncodedAsTuple(_ type: Any) -> Bool {
         let typeStr = "\(type)"
         let pattern = "^\\(Optional\\(Swift\\.String\\), Optional\\(Swift\\.[a-zA-Z0-9]+\\)\\)$"
-        if let regex = try? NSRegularExpression(pattern: pattern, options: []) {
+        do {
+            let regex = try NSRegularExpression(pattern: pattern, options: [])
```
ianpartridge: Because this is creating a fixed, known valid regular expression, I think it would be fine to `try!` actually. If this fails, Foundation is broken.

Sources/Kitura/SwaggerGenerator.swift
```diff
@@ -600,20 +614,25 @@ struct SwaggerDocument: Encodable {
...
             guard self.definitions[model] == nil else {
-                Log.debug("Already generated model \(model)")
+                Log.debug("Already generated model '\(model)'")
                 return
             }
 
             // then build all it
-            if let modelInfo = try? buildModel(typeInfo) {
-                Log.debug("in addModel(model: \(model))")
+            do {
```
ianpartridge: I prefer the `do {}` block to be tightly scoped to the relevant `try`.

Sources/Kitura/SwaggerGenerator.swift
```diff
@@ -600,20 +614,25 @@ struct SwaggerDocument: Encodable {
...
                 if modelInfo.required.count > 0 {
                     modelDefinition = SwaggerModel(type: "object", properties: modelInfo.properties, required: Array(modelInfo.required))
                 } else {
                     modelDefinition = SwaggerModel(type: "object", properties: modelInfo.properties, required: [])
                 }
                 self.definitions[model] = modelDefinition
+            } catch {
+                Log.error("Failed to build model '\(model)': \(error)")
             }
+        } else {
```
ianpartridge: Can we make the `if case` for this a `guard` so we can remove the indentation?

djones6:
Needs https://github.com/IBM-Swift/TypeDecoder/pull/13 to be merged + tagged, as the log formatting improvements depend on the new `TypeDecodeError` type.

Sources/Kitura/SwaggerGenerator.swift
djones6: Agreed

djones6:
Examples of a `TypeDecodingError` (introduced in https://github.com/IBM-Swift/TypeDecoder/pull/13) formatted by SwaggerGenerator during route registration:

No conformance to `ValidKeyedCodingValueProvider` or `ValidSingleCodingValueProvider`:
> <code>[2018-10-11T13:59:03.800+01:00] [WARNING] [SwaggerGenerator.swift:893 logTypeDecodeError(_:for:on:as:)] Output type `ContainsValidations` on path `/testTypeSingular` could not be decoded: TypeDecoder was unable to decode type `YoungAdult`: conformance to `ValidKeyedCodingValueProvider` or `ValidSingleCodingValueProvider` may be required. Underlying error: dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: "Age 0 is outside the permitted range of 18-30", underlyingError: nil))</code>

`ValidKeyedCodingValueProvider` implemented, but decoding still failed:
> <code>[2018-10-11T13:50:36.783+01:00] [WARNING] [SwaggerGenerator.swift:893 logTypeDecodeError(_:for:on:as:)] Output type `ContainsValidations` on path `/testTypeSingular` could not be decoded: TypeDecoder was unable to decode keyed type `YoungAdult`, confirm that `validCodingValue(forKey:)` provides appropriate values for all validated fields. Underlying error: dataCorrupted(Swift.DecodingError.Context(codingPath: [], debugDescription: "Age 12345 is outside the permitted range of 18-30", underlyingError: nil))</code>

Sources/Kitura/SwaggerGenerator.swift
djones6: This will become a switch in a future PR (we need to recursively add models that are nested within unkeyed / dynamicKeyed cases)

djones6:
@ianpartridge From my perspective this is ready - are you happy to merge?

PR #1343, opened 2018-10-02
Don't enable FileResourceServer if handlers exist on the router.

NocturnalSolutions:
## Description

> nocturnal [7:02 PM]
> I’m just now noticing FileResourceServer. It’s… interesting.
> 
> nocturnal [7:05 PM]
> So every Kitura project has a built-in static file server, and you can’t disable it. Nor can you can’t change the path prefix that triggers it, or the directory of files it serves, so it can’t be leveraged to do something useful.
> 
> ianpartridge [4 days ago]
> Yes, this is what serves up the default splash page. I agree it should be possible to disable this.

Accordingly, these changes implicitly disable FileResourceServer if the Router instance has any elements; e.g., if `.get()` or any of its friends have been used to attach a handler to a route.

## Motivation and Context

- Probably the biggest reason is that it's unexpected behavior for Kitura to be able to serve files when it hasn't explicitly been instructed to do so. (This is largely why I engineered this to automatically disable the FileResourceServer functionality if any handlers exist rather than adding a resourceServerEnabled property or something like that that would require users to change their code to get behavior that they're probably already expecting.)
- Not initializing FileResourceServer when not needed is theoretically a performance and memory usage improvement, though the former may be offset by new checks to see if there are no elements on the router.
- Not being able to disable FileResourceServer is arguably a security risk, though fortunately the code seems to be smart enough to not let you back out of the base resource directory on the file system.

## How Has This Been Tested?

One new test case has been added to ensure that resources are no longer served when the router has elements/handlers. Existing test cases which were testing FileResourceServer were modified to do so with an "empty" Router instance.

## Checklist:

## Other Notes:
- To avoid duplication of code and code purpose, it may be desirable for the FileResourceServer to be replaced with a StaticFileServer, though this PR makes no attempt to go down that road yet.
- I can't believe I've been working with Kitura for like a year now and just noticed this a couple days ago. I'm such a n00b.

### DIFF
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -47,7 +47,7 @@ public class Router {
     fileprivate let kituraResourcePrefix = "/@@Kitura-router@@/"
 
     /// Helper for serving file resources
-    fileprivate let fileResourceServer = FileResourceServer()
+    fileprivate lazy var fileResourceServer = FileResourceServer()
 
     /// Flag to enable/disable access to parent router's params
     private let mergeParameters: Bool
@@ -603,7 +603,7 @@ extension Router : ServerDelegate {
             return
         }
 
-        if  urlPath.hasPrefix(kituraResourcePrefix) {
+        if  elements.isEmpty && urlPath.hasPrefix(kituraResourcePrefix) {
             let resource = String(urlPath[kituraResourcePrefix.endIndex...])
             fileResourceServer.sendIfFound(resource: resource, usingResponse: response)
         } else {
@@ -625,7 +625,7 @@ extension Router : ServerDelegate {
     /// - Parameter response: The `RouterResponse` object used to respond to the
     ///                     HTTP request.
     private func sendDefaultResponse(request: RouterRequest, response: RouterResponse) {
-        if request.parsedURLPath.path == "/" {
+        if elements.isEmpty && request.parsedURLPath.path == "/" {
             fileResourceServer.sendIfFound(resource: "index.html", usingResponse: response)
         } else {
             do {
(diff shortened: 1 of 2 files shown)


ianpartridge:
No objection from me, but I guess technically this is a breaking behaviour change?

NocturnalSolutions:
Ya know, I almost kind of think of it as a bug fix, since it's changing things to behave closer to what's expected - in my opinion, anyway. I suppose it's possible that someone is using custom routes and also still expecting the built-in resource server to work, but that's got to be a minority.

Sources/Kitura/Router.swift
```diff
@@ -625,7 +625,7 @@ extension Router : ServerDelegate {
     /// - Parameter response: The `RouterResponse` object used to respond to the
     ///                     HTTP request.
     private func sendDefaultResponse(request: RouterRequest, response: RouterResponse) {
-        if request.parsedURLPath.path == "/" {
+        if elements.isEmpty && request.parsedURLPath.path == "/" {
```
djones6: I agree with the idea of disabling the built-in resource file server / welcome page, but I'm not sure that doing so strictly in the presence of routes will work for us.

Today, if someone runs `kitura init` then a sample project is created with a `/health` route; from `kitura create` there may also be some example routes.  I wouldn't want to suppress the default welcome page in this scenario.

Perhaps a conscious opt-in or opt-out mechanism might work, possibly in conjunction with this change.

djones6:
On the subject of a mechanism to 'opt in' to the default resource server, perhaps `Router.init(mergeParameters: Bool = false)` could be given an additional parameter such as `serveDefaultLandingPage: Bool = false`.

This could then serve to override the decision on whether to serve the default page (the proposal here).  Projects generated by the `kitura` CLI could use `router = Router(serveDefaultLandingPage: true)`, which could later be disabled (or replaced with simply `Router()`) when appropriate.

NocturnalSolutions:
Okay, good idea. Created a new PR with these changes (since I thought it made more sense to start with a fresh branch rather than bash these changes onto the existing one and I can't seem to figure out how to change the "from" branch of an existing PR).

PR #1344, opened 2018-10-03
FileResourceServer must be explicitly enabled when initializing Router

NocturnalSolutions:
## Description

This is a redo of #1343 based on feedback in that thread. Whereas the code in that PR implicitly disabled the FileResourceServer for a Router if that Router had any elements/handlers, this PR changes Router to add a new optional `withResourceServer` parameter to its initializer which only starts the FileResourceServer if explicitly set to `true`. 

## Motivation and Context

@djones6 recommended this approach rather than the one I had chosen in the original PR, and I like his idea better than mine too.  As stated there, this may break expected functionality with the Kitura CLI tool.

## How Has This Been Tested?

As with the other PR, existing tests testing the FileResourceServer functionality have been tweaked, and a new test has been added to ensure the functionality is not present by default.

## Checklist:

### DIFF
--- a/Sources/Kitura/Router.swift
+++ b/Sources/Kitura/Router.swift
@@ -47,7 +47,7 @@ public class Router {
     fileprivate let kituraResourcePrefix = "/@@Kitura-router@@/"
 
     /// Helper for serving file resources
-    fileprivate let fileResourceServer = FileResourceServer()
+    fileprivate let fileResourceServer: FileResourceServer?
 
     /// Flag to enable/disable access to parent router's params
     private let mergeParameters: Bool
@@ -88,9 +88,14 @@ public class Router {
     /// ```
     /// - Parameter mergeParameters: Optional parameter to specify if the router should be able to access parameters
     ///                                 from its parent router. Defaults to `false` if not specified.
-    public init(mergeParameters: Bool = false) {
+    /// - Parameter useResourceServer: Optional parameter to start and use the
+    ///                                FileResourceServer to serve the default
+    ///                                "Welcome to Kitura" page and related
+    ///                                assets.
+    public init(mergeParameters: Bool = false, useResourceServer: Bool = false) {
         self.swagger = SwaggerDocument()
         self.mergeParameters = mergeParameters
+        fileResourceServer = useResourceServer ? FileResourceServer() : nil
 
         Log.verbose("Router initialized")
     }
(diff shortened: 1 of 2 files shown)


Sources/Kitura/Router.swift
```diff
@@ -88,9 +88,14 @@ public class Router {
     /// ```
     /// - Parameter mergeParameters: Optional parameter to specify if the router should be able to access parameters
     ///                                 from its parent router. Defaults to `false` if not specified.
-    public init(mergeParameters: Bool = false) {
+    /// - Parameter useResourceServer: Optional parameter to start and use the
+    ///                                FileResourceServer to serve the default
+    ///                                "Welcome to Kitura" page and related
+    ///                                assets.
+    public init(mergeParameters: Bool = false, useResourceServer: Bool = false) {
```
djones6: Please can we name this param `enableWelcomePage`?  I think it is more descriptive (also the 'ResourceServer' is an internal implementation detail).

Also, having considered it more, the default should be `true`, to make this an opt-out rather than an opt-in.  We want to maintain the existing behaviour by default, otherwise this would have to wait until Kitura 3.  (we could consider turning it off at that time).

Tests/KituraTests/TestStaticFileServer.swift
```diff
@@ -34,6 +34,7 @@ class TestStaticFileServer: KituraTest {
             ("testGetWithWhiteSpaces", testGetWithWhiteSpaces),
             ("testGetWithSpecialCharacters", testGetWithSpecialCharacters),
             ("testGetWithSpecialCharactersEncoded", testGetWithSpecialCharactersEncoded),
+            ("testNoResourcesByDefault", testNoResourcesByDefault),
```
djones6: as per the previous comment, this would become `testWelcomePageCanBeDisabled`

NocturnalSolutions:
Okay, tweaks made and pushed. But yes, I'm hoping this latest commit can be reverted for Kitura 3, as as I've said before, the current behavior is arguably a bug.

djones6:
@NocturnalSolutions this looks good, thanks.  I'm pushing a small change to reverse the order in which we check whether there's a FileResourceServer vs. checking the path, since the latter is more expensive and we can skip it if the welcome page is disabled.

TaborKelly:
Is there a timeline on when this will be available for a released version of Kitura?

ianpartridge:
Perhaps next week, or maybe the week after.

Issue #1337, opened 2018-09-18
KituraTests.TestCustomCoders testCustomQueryEncoder failure


weissi:
## Context and Description

Tests `KituraTests.TestCustomCoders testCustomQueryEncoder` failed

```
Test Case '-[KituraTests.TestCustomCoders testCustomQueryEncoder]' started.
[VERBOSE] [Router.swift:95 init(mergeParameters:)] Router initialized
/private/tmp/.nio-release-tools_dOgew6/Kitura/Tests/KituraTests/TestCustomCoders.swift:156: error: -[KituraTests.TestCustomCoders testCustomQueryEncoder] : XCTAssertEqual failed: ("Optional("&date=2018-02-21T9:47:36%2B0000")") is not equal to ("Optional("&date=2018-02-21T09:47:36%2B0000")") - 
/private/tmp/.nio-release-tools_dOgew6/Kitura/Tests/KituraTests/TestCustomCoders.swift:156: error: -[KituraTests.TestCustomCoders testCustomQueryEncoder] : XCTAssertEqual failed: ("Optional("&date=2018-02-21T9:47:36%2B0000")") is not equal to ("Optional("&date=2018-02-21T09:47:36%2B0000")") - 
/private/tmp/.nio-release-tools_dOgew6/Kitura/Tests/KituraTests/TestCustomCoders.swift:180: error: -[KituraTests.TestCustomCoders testCustomQueryEncoder] : XCTAssertEqual failed: ("Optional("&date=2018-02-21T9:47:36%2B0000")") is not equal to ("Optional("&date=2018-02-21T09:47:36%2B0000")") - 
/private/tmp/.nio-release-tools_dOgew6/Kitura/Tests/KituraTests/TestCustomCoders.swift:180: error: -[KituraTests.TestCustomCoders testCustomQueryEncoder] : XCTAssertEqual failed: ("Optional("&date=2018-02-21T9:47:36%2B0000")") is not equal to ("Optional("&date=2018-02-21T09:47:36%2B0000")") - 
Test Case '-[KituraTests.TestCustomCoders testCustomQueryEncoder]' failed (0.050 seconds).
```

it seems to be a leading 0 that might or might not be there: `T9:` vs `T09:`

### Environment Details

- macOS 10.14
- Xcode 10
- with and without `export KITURA_NIO=1`

### Steps to Reproduce
1) check out current master (`bb19ec7ef8321082a234624cc76c11a17cd88971`)
2) `swift test`

[...]

PR #1349, opened 2018-10-22
Test: CustomCoder test compares dates instead of Strings

Andrew-Lees11:
This PR fixes issue #1337. This issue was that when the custom coders test was comparing the String representation of dates it was saying they were not equal because the strings were different when the dates they represented was equal.

This PR converts the Strings into a Date object and compares these two date objects so that equal dates will be correctly recognized as equal and the test will pass.


Tests/KituraTests/TestCustomCoders.swift
```diff
@@ -152,8 +154,13 @@ class TestCustomCoders: KituraTest {
         
         performServerTest(customRouter) { expectation in
             self.performRequest("get", path: "/rawget", callback: { response in
-                if let response = response, let responseString = try? response.readString() {
-                    XCTAssertEqual(responseString, "&date=2018-02-21T09:47:36%2B0000")
+                if let response = response,
+                   let responseString = try? response.readString(),
+                   let unwrappedString = responseString
+                {
+                    let responseDate = self.dateFormatter.date(from: String((unwrappedString.dropFirst(6))))
```
djones6: Could you add a comment explaining why we are dropping the first 6 characters? (presumably, `&date=`)

Andrew-Lees11: Have added a comment for why we drop the first 6 characters.

PR #1350, opened 2018-10-22
Made TypeSafeMiddleware initialize first in codable routes

Andrew-Lees11:
## Description
This PR changes the TypeSafeMiddleware handlers so that the TypeSafeMiddleware is initialized before, Query parameters, Identifiers and Codable objects. 

## Motivation and Context
This change was requested for two reasons:
- So that you can have a TypeSafeMiddleware that logs Query parameters even when the query fails to be decoded
- So that you can have a TypeSafeMiddleware that altered the request before it is decoded into a Codable object (E.g. Changing the request decoder that will be used)

## How Has This Been Tested?
The existing TypeSafeMiddleware tests have been run and are all still passing.


Andrew-Lees11:
@djones6 
I have added a test case for logging the query parameters using a typeSafeMiddleware even if they are incorrect. With changing decoder, since we are no longer making them public this is no longer a use case.

PR #1351, opened 2018-10-23
Feat: Make request/response decoder/encoder public

Andrew-Lees11:
## Description
This pull request makes the encoders and decoder on the request/response public. This allows the user to set the encoder/decoder for a specific request/response either inside the route for Raw routing or using a Middleware with Codable routing.

## Motivation and Context
This change was requested so that you could create a JWT decoder that selects the public key for decryption based on a header in the request.

## How Has This Been Tested?
A test has been added that sets a custom encoder/decoder inside a Raw route to change the Date decoding strategy. 

## Checklist:

### DIFF
--- a/Sources/Kitura/RouterRequest.swift
+++ b/Sources/Kitura/RouterRequest.swift
@@ -42,11 +42,11 @@ public class RouterRequest {
 
     /// The server request.
     let serverRequest: ServerRequest
-    
-    /// The Data decoder generator for the request content-type
-    let decoder: BodyDecoder?
 
     // MARK: Properties
+    
+    /// The BodyDecoder that will be used to decode Data into a Codable type.
+    public var decoder: BodyDecoder?
     
     /// The hostname of the request.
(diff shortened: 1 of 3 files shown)


Sources/Kitura/RouterRequest.swift
```diff
@@ -42,12 +42,12 @@ public class RouterRequest {
...
     /// The server request.
     let serverRequest: ServerRequest
-    
-    /// The Data decoder generator for the request content-type
-    let decoder: BodyDecoder?
 
     // MARK: Properties
     
+    /// The BodyDecoder that will be used to decode Data into a Codable type.
+    public var decoder: BodyDecoder?
```
djones6: These properties only really need to be publicly settable - would it be more appropriate to keep them internal, but provide a public setter?

Andrew-Lees11:
Closing this PR since the original use case of Changing the JWT coders depending of the key can be solved within JWT class itself.

PR #1353, opened 2018-11-02
[Fix] StaticFileServer request path extraction

ryanpeggs:
## Context and Description
When StaticFileServer was added as a middleware, and it was added not to the root router but to a sub-router, and it had redirect folders to trailing slash turned on (default setting) then its redirection would be incorrect.

### Environment Details
Reproduced Swift 4.0.3 & 4.2; MacOS and Linux

### Steps to Reproduce
1) Setup a simple Kitura app with a router
2) Add a sample HTML file relative to the project at ./src/users/user_id/settings/index.html
3) Add a StaticFileServer to a sub-router of the main router

router.route("/users/:user_id").get(middleware: StaticFileServer(path: "./src/users/user_id/"))

4) HTTP GET /users/me/settings

Expected to happen... 

- Redirect issued to /users/me/settings/ and then the html file loaded

Actually happened... 

- Redirect issued to /settings/ which does not exist (404)

### Fix
Extract path of request from "request.parsedURL" instead of "request.parsedURLPath" as this takes into account the whole request path.

### DIFF
--- a/Sources/Kitura/staticFileServer/StaticFileServer.swift
+++ b/Sources/Kitura/staticFileServer/StaticFileServer.swift
@@ -141,7 +141,7 @@ open class StaticFileServer: RouterMiddleware {
             return
         }
 
-        guard let requestPath = request.parsedURLPath.path else {
+        guard let requestPath = request.parsedURL.path else {
             return
         }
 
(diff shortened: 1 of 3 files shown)


ryanpeggs:
@kilnerm Added tests. Will fix the CLA issue later (github seems to be having issues right now and wont let me access settings to add my other email to my account).

PR #1358, opened 2018-11-14
Cache URLParser objects for reuse

pushkarnk:
Cache URLParser objects for reuse

## Description

URLParser objects are created once in every RouterRequest. Their creation involves a conversion of String representation of a URL to Data followed by a parsing step. Typically, a webservice consists of a limited number of routes, and hence a limited number of target URLs. It may hence make sense to cache the URLParser objects against the String representations of the URLs. Since we are talking about a limited number of URLs here, the possibility of a leak is negligible. 

## Motivation and Context
Caching can lead to a significant performance gain in certain workloads.

## How Has This Been Tested?
Tested on Linux and macOS

### DIFF
--- a/Sources/Kitura/RouterRequest.swift
+++ b/Sources/Kitura/RouterRequest.swift
@@ -110,12 +110,15 @@ public class RouterRequest {
     private var _parsedURL: URLParser?
     internal let parsedURLPath: URLParser
 
+    private static var urlParsers: [String: URLParser] = [:]
+
     /// The parsed URL.
     public private(set) lazy var parsedURL: URLParser = { [unowned self] in
         if let result = self._parsedURL {
             return result
         } else {
-            let result = URLParser(url: self.serverRequest.urlURL.absoluteString.data(using: .utf8)!, isConnect: false)
+            let urlString = self.serverRequest.urlURL.absoluteString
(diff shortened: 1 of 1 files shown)


shmuelk:
I don't understand how you are caching UrlParsers based on URLs.

Won't they often be different? Even if only by an id in the URL.

pushkarnk:
Hi @shmuelk - I am hoping to cache the [KituraNet.URLParser](http://ibm-swift.github.io/Kitura-net/Classes/URLParser.html) objects which simply look like a collection of parsed components of a URL. I assume identical URLs will can share  URLParser objects?

shmuelk:
You are correct that the same exact URLs can share the same URLParser object. However, that will negate your assumption of the number of URLParse objects involved. You will be potentially caching one URLParse object for every single HTTP request ever made to the server. The question will be how many of the URLs in the requests will be duplicates.

pushkarnk:
Thanks @shmuelk I get your question now. Give me some time! I need to do some more work on 
https://github.com/IBM-Swift/Kitura-NIO/pull/106 to answer this question :)

pushkarnk:
After careful consideration, we've dropped the idea of reusing HTTPServerRequest and HTTPServerResponse objects in Kitura-NIO. With that, this proposal also goes out of scope. Closing the PR.

Issue #1368, opened 2018-11-15
KituraTests.TestServer testServerStartStop causes data race


weissi:
## Context and Description

I ran the Kitura tests with Kitura-NIO enabled under TSan. Basically

    KITURA_NIO=1 swift test --sanitize=thread

and hit a problem in `KituraTests.TestServer testServerStartStop`:

```
Test Case '-[KituraTests.TestServer testServerStartStop]' started.

[VERBOSE] [Router.swift:100 init(mergeParameters:enableWelcomePage:)] Router initialized
==================
WARNING: ThreadSanitizer: Swift access race (pid=10632)
  Read of size 8 at 0x000114162440 by thread T2:
    #0 static Kitura.start() Kitura.swift:137 (KituraPackageTests:x86_64+0x9c5aa)
    #1 static Kitura.run() Kitura.swift:115 (KituraPackageTests:x86_64+0x9bed5)
    #2 closure #1 in TestServer.testServerRun() TestServer.swift:141 (KituraPackageTests:x86_64+0x4d95dc)
    #3 thunk for @escaping @callee_guaranteed () -> () <compiler-generated> (KituraPackageTests:x86_64+0x205c20)
    #4 __tsan::invoke_and_release_block(void*) <null>:524288 (libclang_rt.tsan_osx_dynamic.dylib:x86_64h+0x68beb)
    #5 _dispatch_client_callout <null>:524288 (libdispatch.dylib:x86_64+0x3dca)

  Previous modifying access of Swift variable at 0x000114162440 by main thread:
    #0 static Kitura.stop(unregister:) Kitura.swift:174 (KituraPackageTests:x86_64+0x9e4f0)
    #1 closure #2 in TestServer.testServerRun() TestServer.swift:145 (KituraPackageTests:x86_64+0x4d9714)
    #2 thunk for @escaping @callee_guaranteed (@guaranteed Error?) -> () <compiler-generated> (KituraPackageTests:x86_64+0x26054d)
    #3 -[XCTestCase(AsynchronousTesting) waitForExpectationsWithTimeout:handler:] <null>:524288 (XCTest:x86_64+0x17f9c)
    #4 @objc TestServer.testServerRun() <compiler-generated> (KituraPackageTests:x86_64+0x4d9ad0)
    #5 __invoking___ <null>:524288 (CoreFoundation:x86_64h+0x4af8b)

[...]

PR #1377, opened 2018-12-10
Fix: Add locking for Kitura servers

Andrew-Lees11:
This pull request uses NSLock() to make accessing the static httpServerAndPosts and fastCGIServersAndPorts atomic. 
This makes the creation and deletion of Kitura servers thread safe are requested in issue #1368 
This has been tested by running:
```
KITURA_NIO=1 swift test --sanitize=thread
```
Which used to report 1 warning and now reports none.

### DIFF
--- a/Sources/Kitura/Kitura.swift
+++ b/Sources/Kitura/Kitura.swift
@@ -170,12 +170,12 @@ public class Kitura {
             Log.verbose("Stopping HTTP Server on port \(port)...")
             server.stop()
         }
-        
+
         for (server, port) in fastCGIServersAndPorts {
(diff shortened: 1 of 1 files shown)


Sources/Kitura/Kitura.swift
```diff
@@ -159,23 +165,27 @@ public class Kitura {
...
         if unregister {
             httpServersAndPorts.removeAll()
             fastCGIServersAndPorts.removeAll()
         }
+        serverLock.unlock()
     }
 
     typealias Port = Int
+    
+    static private let serverLock = NSLock()
```
ianpartridge: `private static let`

PR #1381, opened 2018-12-25
Maintain the URL query when static file server redirects

ryanpeggs:
## Description
Maintain the URL query when static file server redirects to add trailing / to directory

## Motivation and Context
Ensure that when loading HTML file that cares about query parameters loading with and without / has same effect.

## How Has This Been Tested?
- Tested on local project, should not impact other areas of code

## Checklist:
- [N/A] If applicable, I have updated the documentation accordingly.
- [N/A] If applicable, I have added tests to cover my changes - unsure how to check url after redirect to implement tests.


djones6:
@ryanpeggs Thank you for the PR.  I'm not familiar with the use of query parameters in this scenario - could you include an example to illustrate why this is required?

My other concern is that this is going to impact performance for all users of StaticFileServer, because [`RouterRequest.parsedURL`](https://github.com/IBM-Swift/Kitura/blob/2.6.0/Sources/Kitura/RouterRequest.swift#L114) is a lazily initialized property, and accessing the query parameters imposes some overhead.

It would be better if we could implement this in a way such that the parsedURL property is only accessed in a situation where it is required.

ryanpeggs:
With the existing Kitura code, loading localhost:8080/file?a=1 redirects to localhost:8080/file/ when trailing slash for directories is turned on (serving ./file/index.html).

This pull request rectifies that so that loading localhost:8080/file?a=1 redirects to localhost:8080/file/?a=1 (maintaining the query parameters initially present).

This is useful in our case as we have static HTML/JS/CSS files served for a website. This website has javascript that behaves differently depending on the query parameters. Eg. we have a home page /home that accepts ?welcome=true to show a welcome tutorial. If we load /home?welcome=true then it HTTP redirects to /home/ and our query parameters are gone. The welcome tutorial is not shown. Under the new pull, loading /home?welcome=true redirects to /home/?welcome=true which behaves as expected.

ryanpeggs:
@djones6 RouterRequest.parsedURL is already accessed here: https://github.com/IBM-Swift/Kitura/blob/master/Sources/Kitura/staticFileServer/StaticFileServer.swift#L161 in StaticFileServer.handle, so the lazy value will have already been computed.

djones6:
@ryanpeggs Our process is to merge contributions to master, and then back-port if the changes are needed on a previous release.  At the moment this will be tagged in the next minor (Kitura 2.8) - do you need this back-ported to Kitura 2.6 and/or 2.7?

djones6:

PR #1382, opened 2019-01-01
tests: Use ephemeral ports in Kitura tests

pushkarnk:
Provide an option to run Kitura tests in parallel (`swift test --parallel`) supported by the use of ephemeral ports.

## Description
Kitura tests that fire up a server have the latter listening at a default port of 8080. This stops us from running the tests in parallel because port collisions happen. If we bind the servers to ephemeral ports, it is possible to run the tests in parallel. This pull requests adds support to optionally use ephemeral ports (through the `USE_EPHEMERAL_PORTS` compiler flag). It also adds some travis CI tests that use ephemeral ports and the run in parallel.

## Motivation and Context
Adds an option to run tests in parallel, showcases the use of ephemeral ports, with both KituraNet and KituraNIO

## How Has This Been Tested?
The following test scenarios have been added to travis.yml
1.  Parallel tests(ephemeral ports) with KituraNet on Linux
2.  Parallel tests(ephemeral ports) with KituraNet on macOS
3.  Parallel tests(ephemeral ports) with KituraNIO on Linux
4.  Parallel tests(ephemeral ports) with KituraNIO on macOS

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,7 +31,15 @@ matrix:
     - os: linux
       dist: trusty
       sudo: required
+      env: CUSTOM_TEST_SCRIPT=.swift-test-parallel
+    - os: linux
+      dist: trusty
+      sudo: required
       env: KITURA_NIO=1
+    - os: linux
+      dist: trusty
+      sudo: required
+      env: KITURA_NIO=1 CUSTOM_TEST_SCRIPT=.swift-test-parallel
     - os: linux
       dist: trusty
       sudo: required
@@ -66,7 +74,15 @@ matrix:
     - os: osx
       osx_image: xcode10
       sudo: required
+      env: CUSTOM_TEST_SCRIPT=.swift-test-parallel
+    - os: osx
+      osx_image: xcode10
+      sudo: required
       env: KITURA_NIO=1
+    - os: osx
+      osx_image: xcode10
+      sudo: required
+      env: KITURA_NIO=1 CUSTOM_TEST_SCRIPT=.swift-test-parallel
 
 before_install:
   - git clone https://github.com/IBM-Swift/Package-Builder.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,14 +28,7 @@ matrix:
     - os: linux
       dist: trusty
       sudo: required
-    - os: linux
-      dist: trusty
-      sudo: required
       env: CUSTOM_TEST_SCRIPT=.swift-test-parallel
-    - os: linux
-      dist: trusty
-      sudo: required
-      env: KITURA_NIO=1
     - os: linux
       dist: trusty
       sudo: required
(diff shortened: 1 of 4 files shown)


Tests/KituraTests/KituraTest.swift
```diff
@@ -61,6 +61,16 @@ class KituraTest: XCTestCase {
...
+#if USE_EPHEMERAL_PORTS
+        func shell(_ args: String...) {
+            let task = Process()
+            task.launchPath = "/usr/bin/env"
+            task.arguments = args
+            task.launch()
+            task.waitUntilExit()
+        }
+        shell("./Tests/KituraTests/setup.sh")
+#endif
```
djones6: Shouldn't this be in the singleton init (if we're going to run it from the tests at all)?

I would like to understand why we would want to do this at all though. Shouldn't Travis just run those commands before running the tests? If it's to make local testing easier, then I think it's reasonable to ask the user to do this if it's necessary.

.travis.yml
```diff
@@ -28,10 +28,18 @@ matrix:
     - os: linux
       dist: trusty
       sudo: required
+    - os: linux
+      dist: trusty
+      sudo: required
+      env: CUSTOM_TEST_SCRIPT=.swift-test-parallel
```
djones6: This is good for proving the tests pass, but I think we could replace the serial tests with the parallel ones (otherwise it's an awful lot of builds for Travis).

Tests/KituraTests/KituraTest.swift
```diff
@@ -132,17 +142,21 @@ class KituraTest: XCTestCase {
         }
 
         do {
+#if USE_EPHEMERAL_PORTS
```
djones6: Is there any benefit to having this as an option - why not just use ephemeral ports all the time? We could just have one test that tests we can listen on a specific port, for coverage.

Tests/KituraTests/KituraTest.swift
pushkarnk: Yes, it is for local testing. I didn't want to break the usual `clone-build-test` sequence. We could do away with this by providing instructions for local testing, but again, there's is no easily accessible section in the README to add this.

.travis.yml
pushkarnk: Sure, I will reduce some of the serial tests to parallel ones to make sure we effectively don't add any more CI runs.

Tests/KituraTests/KituraTest.swift
pushkarnk: Again, this is keep the local testing process undisturbed. The `sysctl` commands need superuser access (at least on `macOS`). If we are willing to provide instructions to run the tests locally, we could do away with the compiler flag.

djones6: Do the Kitura tests exhaust the ephemeral port range if we don't do this? By default, ports 49152 - 65535 are available (16k), so I'd be surprised if this wasn't sufficient for a single execution of the tests.

PR #1389, opened 2019-01-18
Log error in StaticFIleServer init if passed a non-directory path

kilnerm:
## Description
This PR updates the StaticFileServer to have a failable initializer, updates the initializer to fail if the `path` parameter is not to a directory and adds a test to ensure the initializer fails when `path` is a file.

## Motivation and Context
Currently the StaticFileServer documentation states the `path` parameter should be a folder but the initializer will accept a file which can lead to files being incorrectly served. This change prevents this scenario.

## How Has This Been Tested?
I have added a test which will try to create a StaticFileServer using a path to a file, should the StaticFileServer be successfully created then the test will fail. I have also stepped through the debugger to ensure the code functions as expected.

## Checklist:
- [x] Make StaticFileServer initializer failable
- [x] Add failing tests
- [x] Update StaticFileServer initializer to fail.

### DIFF
--- a/Sources/Kitura/staticFileServer/StaticFileServer.swift
+++ b/Sources/Kitura/staticFileServer/StaticFileServer.swift
@@ -106,7 +106,8 @@ open class StaticFileServer: RouterMiddleware {
                  customResponseHeadersSetter: ResponseHeadersSetter? = nil) {
         let rootPathAbsolute = StaticFileServer.ResourcePathHandler.getAbsolutePath(for: path)
         absoluteRootPath = rootPathAbsolute
-        // Check the supplied path is a directory and if not fail initialisation
+        // If the supplied path does not exist log a warning as the path could be created dynamically at runtime.
+        // If the supplied path exists and is not a directory then log an error.
         var isDirectory = ObjCBool(false)
 
         let pathExists = FileManager.default.fileExists(atPath: absoluteRootPath, isDirectory: &isDirectory)
@@ -118,7 +119,7 @@ open class StaticFileServer: RouterMiddleware {
         if !pathExists {
             Log.warning("StaticFileServer being initialised with invalid path: \(rootPathAbsolute)")
         } else if !isDirectoryBool {
-            Log.error("StaticFileServer should not be intialised with a path which resolves to a file")
+            Log.error("StaticFileServer should not be initialised with a path that resolves to a file")
         }
 
         let cacheOptions = options.cacheOptions
--- a/Sources/Kitura/staticFileServer/StaticFileServer.swift
+++ b/Sources/Kitura/staticFileServer/StaticFileServer.swift
@@ -102,7 +102,7 @@ open class StaticFileServer: RouterMiddleware {
     /// - Parameter customResponseHeadersSetter: an object of a class that
     /// implements `ResponseHeadersSetter` protocol providing a custom method to set
     /// the headers of the response.
-    public init?(path: String = "./public", options: Options = Options(),
+    public init(path: String = "./public", options: Options = Options(),
                  customResponseHeadersSetter: ResponseHeadersSetter? = nil) {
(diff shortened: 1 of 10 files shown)


Sources/Kitura/staticFileServer/StaticFileServer.swift
```diff
@@ -113,7 +115,10 @@ open class StaticFileServer: RouterMiddleware {
         #else
         let isDirectoryBool = isDirectory
         #endif
-        guard pathExists, isDirectoryBool else {
+        if !pathExists {
```
djones6: :+1: Failable initializer or not, I prefer this approach, because users may have a legitimate reason for creating a server on a non-existent path (which may come into existence during the server's lifespan).

Sources/Kitura/staticFileServer/StaticFileServer.swift
```diff
@@ -101,9 +102,25 @@ open class StaticFileServer: RouterMiddleware {
     /// - Parameter customResponseHeadersSetter: an object of a class that
     /// implements `ResponseHeadersSetter` protocol providing a custom method to set
     /// the headers of the response.
-    public init(path: String = "./public", options: Options = Options(),
+    public init?(path: String = "./public", options: Options = Options(),
```
djones6: This is a breaking change.  For Kitura 2.x, we'll need to leave it as a non-failable initializer, and log at `.error` instead.

Sources/Kitura/staticFileServer/StaticFileServer.swift
```diff
@@ -101,9 +102,25 @@ open class StaticFileServer: RouterMiddleware {
...
+        let pathExists = FileManager.default.fileExists(atPath: absoluteRootPath, isDirectory: &isDirectory)
+        #if !os(Linux) || swift(>=4.1)
+        let isDirectoryBool = isDirectory.boolValue
+        #else
+        let isDirectoryBool = isDirectory
+        #endif
+        if !pathExists {
+            Log.warning("StaticFileServer being initialised with invalid path: \(rootPathAbsolute)")
+        } else if !isDirectoryBool {
+            Log.error("StaticFileServer cannot be intialised with a file")
```
djones6: If the initializer is no longer failable, this becomes 'should not be initialized with a file'

Sources/Kitura/staticFileServer/StaticFileServer.swift
```diff
@@ -103,7 +104,22 @@ open class StaticFileServer: RouterMiddleware {
...
+        let pathExists = FileManager.default.fileExists(atPath: absoluteRootPath, isDirectory: &isDirectory)
+        #if !os(Linux) || swift(>=4.1)
+        let isDirectoryBool = isDirectory.boolValue
+        #else
+        let isDirectoryBool = isDirectory
+        #endif
+        if !pathExists {
+            Log.warning("StaticFileServer being initialised with invalid path: \(rootPathAbsolute)")
+        } else if !isDirectoryBool {
+            Log.error("StaticFileServer should not be intialised with a path which resolves to a file")
```
djones6: Typo: initialised
Also grammar, which -> that

Sources/Kitura/staticFileServer/StaticFileServer.swift
```diff
@@ -103,7 +104,22 @@ open class StaticFileServer: RouterMiddleware {
     /// the headers of the response.
     public init(path: String = "./public", options: Options = Options(),
                  customResponseHeadersSetter: ResponseHeadersSetter? = nil) {
-        absoluteRootPath = StaticFileServer.ResourcePathHandler.getAbsolutePath(for: path)
+        let rootPathAbsolute = StaticFileServer.ResourcePathHandler.getAbsolutePath(for: path)
+        absoluteRootPath = rootPathAbsolute
+        // Check the supplied path is a directory and if not fail initialisation
```
djones6: This comment needs updating

PR #1390, opened 2019-01-20
[DO NOT MERGE] Default to Kitura-NIO for networking

pushkarnk:
Use Kitura-NIO as the default networking library for Kitura

## Description
We currently default to Kitura-net for networking and offer a switch to Kitura-NIO using the KITURA_NIO environment variable. This commit proposes defaulting to Kitura-NIO and switching to Kitura-net using the KITURA_NET environment variable.

## Motivation and Context
We'd like to have Kitura running over SwiftNIO by default. Kitura-net would be optional.

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,7 +34,7 @@ matrix:
       dist: trusty
       sudo: required
       services: docker
-      env: DOCKER_IMAGE=ubuntu:16.04 KITURA_NIO=1 SWIFT_TEST_ARGS="--parallel"
+      env: DOCKER_IMAGE=ubuntu:16.04 KITURA_NET=1 SWIFT_TEST_ARGS="--parallel"
     - os: linux
       dist: trusty
       sudo: required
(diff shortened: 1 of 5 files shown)


pushkarnk:
The failures on macOS are related to a recent change to `Package-Builder`. Raised a new PR to rectify it. https://github.com/IBM-Swift/Package-Builder/pull/157

pushkarnk:
@djones6 Makes sense

ianpartridge:
Closing this for now.

kyleishie:
Is this still up for consideration?  Correct me if I'm wrong but Swift-NIO 2.0.0+ solves the concerns mentioned above.

djones6:
@kyleishie Swift-NIO 2 requires a minimum of Swift 5, and we intend to continue support for Swift 4 in the Kitura 2.x releases.  It _would_ be possible to make NIO the default when compiling with Swift 5 or higher.  @ianpartridge / @pushkarnk have we considered this?

ianpartridge:
I don't think we've discussed that. Personally I think this is a Kitura 3 item.

PR #1395, opened 2019-01-30
Use UTF8View to create Data

ianpartridge:
A few places in Kitura we use `str.data(using: .utf8)` to create a `Data`. Currently in Swift 5 this is not optimised and goes via a UTF-16 `NSString`.

Instead, use the `Data` initializer which accepts a `Sequence`. This has a fastpath for UTF-8 already.

### DIFF
--- a/Sources/Kitura/bodyParser/MultiPartBodyParser.swift
+++ b/Sources/Kitura/bodyParser/MultiPartBodyParser.swift
@@ -19,24 +19,13 @@ import LoggerAPI
 
 class MultiPartBodyParser: BodyParserProtocol {
     let boundaryData: Data
-    let endBoundaryData: Data
-    let newLineData: Data
-    let endHeaderData: Data
+    let endBoundaryData = Data("--".utf8)
+    let newLineData = Data("\r\n".utf8)
+    let endHeaderData = Data("\r\n\r\n".utf8)
 
     init(boundary: String) {
-        guard let boundaryData = String("--" + boundary).data(using: .utf8),
-            let endBoundaryData = "--".data(using: .utf8),
-            let newLineData = "\r\n".data(using: .utf8),
-            let endHeaderData = "\r\n\r\n".data(using: .utf8)
-            else {
-                Log.error("Error converting strings to data for multipart parsing")
-                exit(1)
-        }
-
-        self.boundaryData = boundaryData
-        self.endBoundaryData = endBoundaryData
-        self.newLineData = newLineData
-        self.endHeaderData = endHeaderData
+        let str = "--" + boundary
+        self.boundaryData = Data(str.utf8)
     }
 
     func parse(_ data: Data) -> ParsedBody? {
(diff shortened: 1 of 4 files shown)


djones6:
Perf results on Linux (16.04) using QueryParams benchmark with `/echo?message=foo`
```
               | Throughput (req/s)      | CPU (%) | Mem (kb)     | Latency (ms)                   | good
Implementation | Average    | Max        | Average | Avg peak RSS | Average  | 99%      | Max      | iters
---------------|------------|------------|---------|--------------|----------|----------|----------|-------
      base_421 |    31237.7 |    31962.7 |    98.0 |        20264 |      4.1 |      5.3 |     24.3 |     5
      fast_421 |    32125.8 |    32400.1 |    98.4 |        20784 |      3.9 |      4.4 |     22.6 |     5
       base_50 |    30008.3 |    30509.1 |    98.6 |        22989 |      4.2 |      4.7 |     27.3 |     5
       fast_50 |    30455.7 |    30791.2 |    98.6 |        17180 |      4.2 |      4.8 |     34.0 |     5
```
Small improvement in the 1.5 - 3% range.

Sources/Kitura/bodyParser/MultiPartBodyParser.swift
```diff
@@ -19,24 +19,13 @@ import LoggerAPI
 
 class MultiPartBodyParser: BodyParserProtocol {
     let boundaryData: Data
-    let endBoundaryData: Data
-    let newLineData: Data
-    let endHeaderData: Data
+    let endBoundaryData = Data("--".utf8)
+    let newLineData = Data("\r\n".utf8)
+    let endHeaderData = Data("\r\n\r\n".utf8)
```
djones6: could these be static?

Sources/Kitura/staticFileServer/FileServer.swift
```diff
@@ -236,11 +236,11 @@ extension StaticFileServer {
                     partHeader += "Content-Range: bytes \(range.lowerBound)-\(range.upperBound)/\(fileSize)\r\n"
                     partHeader += (contentType == nil ? "" : "Content-Type: \(contentType!)\r\n")
                     partHeader += "\r\n"
-                    data.append(partHeader.data(using: .utf8)!)
+                    data.append(Data(partHeader.utf8))
                     data.append(fileData)
-                    data.append("\r\n".data(using: .utf8)!)
+                    data.append(Data("\r\n".utf8))
```
djones6: could this be a constant?

Sources/Kitura/bodyParser/MultiPartBodyParser.swift
ianpartridge: Yes but then they have to be referenced as `MultiPartBodyParser.endBoundaryData` everywhere which makes the diff bigger and the code uglier. What do you think?

Sources/Kitura/staticFileServer/FileServer.swift
ianpartridge: It could, we can add that in future.

Sources/Kitura/bodyParser/MultiPartBodyParser.swift
ianpartridge: Decided leave this as-is for now.

Issue #1388, opened 2019-01-17
Swift 5 performance opportunities


djones6:
Look for performance opportunities with Kitura running on Swift 5 (both with -net and -NIO), and prototype improvements.

## Context and Description
In #1372, we performed a basic performance assessment of Kitura with Kitura-net with Swift 5, and fixed a regression (IBM-Swift/Kitura-net#286), achieving parity with 4.2.

There are likely to be other opportunities to improve performance, including:
- Changes to the internals of String and the relative efficiency of converting to/from utf8-encoded Data
- Data itself has also improved since various parts of Kitura and Kitura-net were originally implemented, and our usage of Data (or lack thereof) is well overdue for a review.

djones6:
As String is now stored internally as utf8, bridging to NSString represents a significant overhead in Swift 5.  There are many places that String is bridged to NSString, which I recall have featured in our code in the past.  We should avoid these on critical paths if possible:

String initializers:
- `init(from: Data)`: https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSStringAPI.swift#L368
  - stdlib fast path: `init(decoding: Collection, as: Unicode.Encoding)` when the encoding is `UTF8`
- similarly from a C string, with contentsOf: URL and contentsOfFile: String

Obtaining Data or bytes:
- `String.data(using:)`: https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSStringAPI.swift#L766
- `String.cString(using:)`: https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSStringAPI.swift#L751
- `String.getCString()`: https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSStringAPI.swift#L705
  - stdlib provides access via UnsafePointer: `withCString(encodedAs: Unicode.Encoding)` - fastpath when encoding is `UTF8`: https://github.com/apple/swift/blob/master/stdlib/public/core/String.swift#L401
  - and `String.utf8CString: ContiguousArray<CChar>`: https://github.com/apple/swift/blob/master/stdlib/public/core/StringUTF8View.swift#L265

Transformations:
- `String.components(separatedBy:)`: https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSStringAPI.swift#L705
- `String.lowercased(with: Locale)`: https://github.com/apple/swift-corelibs-foundation/blob/master/Foundation/NSStringAPI.swift#L899
[...]

PR #1401, opened 2019-02-05
Avoid NSString functions in query params

djones6:
## Description

Replaces the use of `String.components` with `String.split` when processing the query parameters string, and replaces the use of `String.replacingOccurrences` with a new function that uses `String.replaceSubrange` instead.

I did attempt to replace the `removingPercentEncoding()` function with a version using only stdlib functions, however I could not get this to outperform the Foundation implementation consistently on Mac and Linux.  For reference, the implementation I came up with (based on CoreFoundation) was:
https://github.com/IBM-Swift/Kitura/compare/ef0e241de858e96e4165b4d4250b16627288488a...177d74eb010261ee194b40f2b2203d00243f0d88

## Motivation and Context

Avoids bridging to NSString.  (see #1388)

## Checklist:

### DIFF
--- a/Sources/Kitura/String+Extensions.swift
+++ b/Sources/Kitura/String+Extensions.swift
@@ -25,8 +25,10 @@ extension String {
     var urlDecodedFieldValuePairs: [String: String] {
         var result: [String: String] = [:]
         for item in self.split(separator: "&") {
-            let (key, value) = item.keyAndDecodedValue
-            if let value = value {
+            let (keySub, valueSub) = item.keyAndDecodedValue
+            if let valueSub = valueSub {
+                let value = String(valueSub)
+                let key = String(keySub)
                 // If value already exists for this key, append it
                 if let existingValue = result[key] {
                     result[key] = "\(existingValue),\(value)"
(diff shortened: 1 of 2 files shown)


Sources/Kitura/String+Extensions.swift
```diff
@@ -44,32 +46,52 @@ extension String {
...
         }
-        return (key: key, value: decodedValue ?? valueReplacingPlus)
+        return (key: key, value: Substring(decodedValue))
     }
 
+    /// Finds and replaces all occurrences of a character with the provided substring
+    /// (eg. another character).
+    @inline(__always)
+    private mutating func replaceCharacters(_ src: Character, with dst: Substring) {
+        repeat {
```
ianpartridge: Can just be `while true { }`

Sources/Kitura/String+Extensions.swift
```diff
@@ -24,9 +24,11 @@ extension String {
     /// values.
     var urlDecodedFieldValuePairs: [String: String] {
         var result: [String: String] = [:]
-        for item in self.components(separatedBy: "&") {
-            let (key, value) = item.keyAndDecodedValue
-            if let value = value {
+        for item in self.split(separator: "&") {
+            let (keySub, valueSub) = item.keyAndDecodedValue
```
ianpartridge: Why not make `keyAndDecodedValue` continue to return `(String, String?)`? There must be a reason you changed it but I couldn't spot it...

ianpartridge: Oh is it because `item` is a substring? If so, why not make it a string before calling `keyAndDecodedValue`?

djones6: This was to save cost, there's no need to may the key a String if the value is nil.

PR #1403, opened 2019-02-08
Update to Swift 4.2.2 and official Docker images

djones6:
## Description

Updates `.swift-version` to 4.2.2, and makes use of the official swift Docker images (https://hub.docker.com/_/swift/) to improve our build times slightly.

Also, Xenial is now supported natively in Travis, so I have updated our Linux jobs to use it, as I expect Trusty will go away at some point in the future.  We could now run directly on Xenial (and download the builds from swift.org as before), but it seems to be slightly faster to pull the Docker images.


.travis.yml
```diff
@@ -18,35 +18,35 @@ branches:
 matrix:
   include:
     - os: linux
-      dist: trusty
+      dist: xenial
       sudo: required
-      env: SWIFT_SNAPSHOT=4.0.3
+      env: DOCKER_IMAGE=swift:4.0.3 SWIFT_SNAPSHOT=4.0.3
```
ianpartridge: Doesn't this need `services: docker`?

djones6: Good spot!  I'm surprised it worked anyway

PR #1404, opened 2019-02-09
Fix crash when empty MediaType is passed in request

diejmon:
## Description
I've been testing my Kitura server and crashed it with:
```
curl -X "POST" "[ENDPOINT_URL]" \
     -H 'Content-Type: ' \
     -d $'{}'
```

## Motivation and Context
It fixes server crash

## How Has This Been Tested?
Just send a request with empty Content-Type.

## Checklist:


ianpartridge:
Thanks!  I'll backport this to the `kitura-2.6` branch and do a release.

djones6:
Note: I found that (with my version of Curl, at least) I had to use this syntax to persuade curl to pass an empty header: `-H 'Content:Type;'`.  Otherwise, the header is omitted completely.  For debugging you can specify `-iv` to see both the request and response headers to confirm you're sending what you expected.

PR #1420, opened 2019-03-15
Bug fixes for SwaggerGenerator

### LINKED ISSUE #1360: Swagger fixes for nested / array types
## Context and Description

Some nested types or cases where an array type is returned in place of a Codable are not correctly modelled by the SwaggerGenerator.

WIP: https://github.com/IBM-Swift/Kitura/tree/issue.swaggerFixes
Test project: https://github.com/djones6/OpenAPIDemo

### LINKED ISSUE #1416: SwaggerGenerator does not reflect Date encoding strategy
## Context and Description
SwaggerGenerator uses the data format reported by the TypeDecoder for all types while documenting API calls.

Foundation's `JSONEncoder` supports multiple `Date` encoding formats, such as with an `iso8601` formatted string.  In https://github.com/IBM-Swift/Kitura/pull/1309 we provided the ability to customize the JSONEncoder including the date encoding strategy.

By default, a `Date` is encoded as a single value, `Double` type: https://github.com/apple/swift-corelibs-foundation/blob/swift-4.2.3-RELEASE/Foundation/Date.swift#L272, and this is what TypeDecoder reports when asked to describe this type.

However, this means that SwaggerGenerator always reports a Date field as having a numeric type, which is incorrect if the Date encoding strategy is a formatted String.

To fix this, we'd need to special case Date, and consult the JSONEncoder registered with the Router to determine its representation.

### LINKED ISSUE #1419: Regression: SwaggerGenerator does not handle get by ID
## Context and Description
#1347 enabled GET (plural) and POST routes that return an id to be represented in swagger - in the GET (plural) case, as `[(Identifier, Model)]`. It however introduced a regression for routes that get _by_ id, such as `/person/{id}`.

### Environment Details
Kitura 2.6.2

### Steps to Reproduce
```swift
    // Get a single person by ID
    app.router.get("/personById") { (id: Int, respondWith: @escaping (Person?, RequestError?) -> Void) in
        Person.find(id: id) { (person, error) in
            guard let person = person else {
                return respondWith(nil, error)
            }
            return respondWith(person, nil)
        }
    }
```

- [ ] Expected to happen...
Swagger (note - simplified) describing this route as:
```json
    "/personById/{id}" : {
      "get" : {
        "parameters" : [ {
            "in" : "path",
            "name" : "id",
            "required" : true,
            "type" : "integer"
        } ],
        "responses" : {
          "200" : {
            "schema" : {
              "$ref" : "#/definitions/Person"
            }
    } } } }
```
- [ ] Actually happened...
```json
    "/personById" : {
      "get" : {
        "responses" : {
          "200" : {
            "schema" : {
              "$ref" : "#/definitions/Person"
            }
    } } } }
```

djones6:
## Description

This PR fixes a number of bugs and shortcomings of SwaggerGenerator:
- Handle custom Date encodings in SwaggerGenerator (resolves #1416) - note, a limitation of the current design of SwaggerGenerator means that custom encoders *must* be configured _before_ routes are registered.
- Handle Codable GET (singular) return types that are Arrays (resolves #1360) - this is not a particularly beautiful solution and relies on stringifying the Swift type and seeing whether it starts with `Array<`, however it's the only solution I could find for now.
- Fix regression for GET (singular) by ID (resolves #1419).  There was a test for this, but that test was (erroneously) modified by #1347 such that it passed.  I've reverted that change so that it tests for the correct behaviour (appending `/{id}` to routes that accept an identifier).

Also fixes several cases that I encountered while experimenting:
- Generate models that only appear as types nested within Array or Dictionary values (which previously were missed, leading to invalid refs).
- Handle models that encode to a single value, complex type, such as struct A encoding itself as a single value of struct B.  Previously, the generator produced a ref to A, but generated a model definition for B.
- Remove resolved FIXMEs and TODOs from Router and SwaggerGenerator.


djones6:
@ianpartridge well, it's a good thing I added tests: I found a bug with the combination of supporting GET(singular) array return types and GET(plural) tuple arrays.

This is now ready for review 😄

Issue #1418, opened 2019-03-12
If application ends with error, the process exit code should be different than 0


krzyzanowskim:
The "Hello World" Kitura application error handling is not optimal. For the case presented below, the executable exit with exit code 0 - that means "success", while there was an error and exit code should be different than 0 (eg. 1).

If application ends with error, the process exit code should be different than 0

```
$ ./Hello
[2019-03-12T09:14:35.698+01:00] [WARNING] [ConfigurationManager.swift:389 load(url:deserializerName:)] Unable to load data from URL /Users/marcinkrzyzanowski/Devel/SwiftStudio/Examples/Kitura/Hello/.build/x86_64-apple-macosx/debug/config/mappings.json
[Tue Mar 12 09:14:35 2019] com.ibm.diagnostics.healthcenter.loader INFO: Swift Application Metrics
[2019-03-12T09:14:35.744+01:00] [INFO] [Metrics.swift:20 initializeMetrics(router:)] Initialized metrics.
[2019-03-12T09:14:35.746+01:00] [INFO] [Application.swift:28 run()] http://localhost:8080/
[2019-03-12T09:14:35.748+01:00] [ERROR] [Kitura.swift:139 start()] Error listening on port 8080: Error code: -9992(0x-2708), Address already in use. Use server.failed(callback:) to handle
$ echo $?
0
```

sympatheticmoose:
related issue - https://github.com/IBM-Swift/Kitura/issues/1191

ianpartridge:
Thanks Marcin, we will take a look ASAP.

djones6:
I agree, this behaviour is unusual.  There are several possible ways we could behave, due to how Kitura starts listeners (there can be more than one), and what `Kitura.run()` does.

For a simple application that contains a single listener socket (typically 8080 or 8090), this is usually followed by invoking `Kitura.run()` on the main thread.

`Kitura.run()` will first invoke `Kitura.start()`, which:
    - tries to create a socket on the specified port, and listen for incoming connections
    - if successful, a background thread is created to accept new connections to the socket.

`Kitura.run()` will then block, waiting for all listeners to complete (indefinitely, or until `Kitura.stop()` is called).

In the case where the port is unavailable, that listener will fail to start.  We currently log an error here, but do not produce any return code, nor do we call `exit()`.

[...]

PR #1430, opened 2019-03-27
Enable Kitura to exit with a non-zero exit code

ddunn2:
## Description
This PR updates the `Kitura.run()` method to the following: 'Kitura.run(withStatus: Bool)'. This enables a user to choose whether they want their application process to end and exit with the correct code. The `withStatus` parameter is given a default value of `false` to prevent this being a breaking change.

## Motivation and Context
Solves issue: [#1418](https://github.com/IBM-Swift/Kitura/issues/1418)

## How Has This Been Tested?

- Ran the Kitura tests
- Test dummy applications


Sources/Kitura/Kitura.swift
```diff
@@ -105,6 +105,7 @@ public class Kitura {
     // MARK: Start Servers
     
     /// Start the Kitura framework.
+    /// Can also provide a value for the withStatus param, this will enable the process to exit with an exit code should any of the servers fail to start.
```
helenmasters: This sentence could do with rewording.
Plus you need to add the `withStatus` parameter and its description to the jazzy doc.

ddunn2: Reworded to what?
I'll look into how to add the `withStatus` parameter to the Jazzy docs, it's been a long time since I had to work with Jazzy. 

Can we add a test for this? Do the Kitura tests allow for us to spin up multiple servers on the same port within within the same test?

helenmasters: Sorry David, more info on wording... I'd expect the parameter doc for `withStatus` to say something like:

`withStatus` "Determines whether the Kitura process can return a non-zero exit code should any of the servers fail to start. Defaults to false, indicating it will not return any non-zero exit codes."

I'd expect the descriptive wording to be something like.
/// Start the Kitura framework.
/// By default, the Kitura framework process will exit with a code of zero, even if one or more of the servers fails to start.  To enable the Kitura framework process to exit with a non-zero exit code set the `withStatus` parameter to true.

ddunn2: Thanks for providing that, it definitely reads better than what I had :). 

I've just pushed up your suggested changes.

Sources/Kitura/Kitura.swift
```diff
@@ -149,6 +164,40 @@ public class Kitura {
...
+    /// Start all registered servers and return the number of servers that failed to start.
+    ///
+    ///### Usage Example: ###
+    /// Make all registered servers start listening on their port.
+    ///```swift
+    /// let router = Router()
+    /// Kitura.addHTTPServer(onPort: 8080, with: router)
+    /// Kitura.startWithStatus() // Returns the number of failed server starts.
+    ///```
+    public class func startWithStatus() -> Int {
```
djones6: Rather than duplicating code, can we make the existing `start()` call this one, eg:
```swift
public class func start() {
    _ = startWithStatus()
}
```

Sources/Kitura/Kitura.swift
```diff
@@ -113,10 +114,24 @@ public class Kitura {
...
+    /// Make all registered servers start listening on their port and exit if any fail to start.
+    ///```swift
+    /// let router = Router()
+    /// Kitura.addHTTPServer(onPort: 8080, with: router)
+    /// Kitura.run(withStatus: true)
+    ///```
     /// - note: This function never returns - it should be the last call in your `main.swift` file.
-    public class func run() {
+    /// - Parameter withStatus: Determines whether the Kitura process can return a non-zero exit code should any of the servers fail to start. Defaults to false, indicating it will not return any non-zero exit codes.
+    public class func run(withStatus status: Bool = false) {
```
djones6: I don't like the name `withStatus` here, as this function doesn't return an exit code, it calls `exit` if a failure occurs.  How about `exitOnFailure`?

I also think there should be a `run()` variant that does return the status code, eg. `runWithStatus() -> Int`, so that a user has the option of handling a situation where one or more servers failed to start.  This function should have a parameter that controls whether run waits if at least one listener is successful, or if it should return immediately if at least one failure occurs.

Sources/Kitura/Kitura.swift
```diff
@@ -105,6 +105,7 @@ public class Kitura {
     // MARK: Start Servers
     
     /// Start the Kitura framework.
+    /// By default, the Kitura framework process will exit with a code of zero, even if one or more of the servers fails to start. To enable the Kitura framework process to exit with a non-zero exit code set the `withStatus` parameter to true.
```
djones6: By default, the Kitura framework does not call `exit` at all - the reason a Kitura process typically exits with 0, is that the `run()` function completes (as no servers remain to listen on), and `Kitura.run()` is typically the final line of code for the main thread to execute.  Once the last line completes, the default action is that the process ends with a return code of zero.

If the user is able to determine that there was an error, then they could call `exit` themselves with a non-zero code, ie:
```swift
let numberOfFailures = Kitura.runWithStatus()
exit(numberOfFailures)
```

Sources/Kitura/Kitura.swift
ianpartridge: I'm not sure about calling `exit()` at all... Can we return a `@discardableResult` from `run()` instead and let the user decide what to do?

djones6: @ianpartridge if we add an `@discardableResult` to the existing `run()` then the function no longer has the same return type and would be a breaking change.  Now granted, this is _unlikely_ to break users, but it would if someone wrote this:
```swift
func startMyServer() -> Void {
  return Kitura.run()
}
```
Unfortunately you can't overload `run()` by return type only.  Well, you _can_, but that's also breaking because people who just write `Kitura.run()` will then get a compilation failure because the statement is ambiguous.

djones6:
@ddunn2 Can you address the latest review comments?

Sources/Kitura/Kitura.swift
ddunn2: I don't like the parameter name `withStatus` either but it's what was suggested in the issue so I used it as I couldn't decide on anything better. 

I don't really like `exitOnFailure ` either. What about:

`run(withExit: true)`
`run(asExitable: true)`
`run(inExitMode: true)`

Or something like that, basically something that reads `Swift-like`. 

Rather then adding another public function, could be not add a property that keeps track of how many failures occurred? 

Could look something like this:

```swift
run(withStatus = true, canFail = true) //Needs a better name
```
Then inside `startWithStatus` we can check whether those values have been set and respond appropriately. 

- If `true, true`:  return immediately if at least one failure occurs.
- if `true, false`: wait to see if at least one listener is successful.
- if `false, any`: Can't exit so ignore second param. 

Just an idea, I haven't tested it yet nor am I aware of any implications of implementing it this way.

ianpartridge: `exitOnFailure`?

ianpartridge: Oh Dave already suggested it! Sounds least-worst to me...

djones6:
Updated proposal:
- `run()` takes a new `exitOnFailure: Bool` which defaults to `true`
- `startWithStatus()` is added, returning a type which indicates whether there was failure (could be an Int count of fails, or something more descriptive)
- `start()` just calls `startWithStatus()` and discards the result
- `wait()` is implemented, which does what `run()` does, so that people could (for example):
```swift
let failures = Kitura.startWithStatus()
if failures == 0 {
  Kitura.wait()
} else {
  // handle failure
}
```

This avoids us having to implement `runWithStatus()` as well as `startWithStatus()`, and the default behaviour of `run()` then fixes the original issue that Kitura exits with 0 if you do something simple such as add a single server and call run(), but the port is already in use.

Sources/Kitura/Kitura.swift
```diff
@@ -155,10 +156,24 @@ public class Kitura {
     /// Kitura.addHTTPServer(onPort: 8080, with: router)
     /// Kitura.run()
     ///```
+    /// Make all registered servers start listening on their port and exit if any fail to start.
+    ///```swift
+    /// let router = Router()
+    /// Kitura.addHTTPServer(onPort: 8080, with: router)
+    /// Kitura.run(exitOnFailure: false)
```
djones6: @ddunn2 I just spotted that the docs here are incorrect - since the default for `exitOnFailure` is `true`, the first example needs to set it to false, and the second example can use the default.  Or, swap the titles of the examples so that the first one (the default) is exit if there's a failure to start.

djones6:
I've updated the opening comment for this PR to more accurately reflect the new API.

djones6:

PR #1440, opened 2019-04-12
Add the NIO switch to the 4.x.x Package.swift files

pushkarnk:
The Kitura-NIO switch needs to be supported on 4.x.x versions of the manifest file.

## Description
PR https://github.com/IBM-Swift/Kitura/pull/1436 removed support for KITURA_NIO. I think this was based on the decision that we'd be supporting Kitura-NIO only with Swift 5 (and SwiftNIO 2). However, we now have a `kitura-nio-1.0` branch on `Kitura-NIO` and we run Kitura tests in the CI for this. This needs Kitura-NIO support for Kitura with swift 4.x.x too.

### DIFF
--- a/Package@swift-4.0.swift
+++ b/Package@swift-4.0.swift
@@ -19,6 +19,14 @@
 import PackageDescription
 import Foundation
 
+var kituraNetPackage: Package.Dependency
+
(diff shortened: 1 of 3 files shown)


pushkarnk:
cc @ianpartridge @djones6

PR #1441, opened 2019-04-15
Update/add font content types

cgossain:
Specifically, the changes are as follows:

[Updated] 			.otf (application/x-font-otf -> font/otf)
[Removed Duplicate] 	.otf (font/opentype)
[Added] 				.sfnt (font/sfnt)
[Updated] 			.ttf/ttc (application/x-font-ttf -> font/ttf)
[Updated] 			.woff (application/font-woff -> font/woff)
[Updated] 			.woff2 (application/font-woff2 -> font/woff2)

All other font definitions have been left intact. These changes were based on this Stack Overflow question/answer:
https://stackoverflow.com/a/20723357/485352

## Checklist:


ianpartridge:
Hi @cgossain thanks for your PR!  The content types here are actually taken from the `node-mime` project:

https://github.com/broofa/node-mime/blob/master/types/standard.json

It looks like they have already made this fix "upstream", so I think what we should do is take the latest version of this array from `node-mime` which I suspect will include other fixes as well as the font ones.

Would you mind doing that?

cgossain:
Hey @ianpartridge, yes I can absolutely do that, I'll try and get that done later today.

cgossain:
@ianpartridge Hey Ian, the `standard.json` file looks a lot less comprehensive than what's included in Kitura. Can I assume that what's included in Kitura is a combination of `standard.json` + `other.json`?

https://github.com/broofa/node-mime/blob/master/types/other.json

ianpartridge:
Hmm! Yes I think so. Perhaps `node-mime` only had one large file when we borrowed it :) Can we merge them?

cgossain:
@ianpartridge Probably. To keep everything in alphabetical order, I can create a quick little swift based command line tool that would take the two files as inputs and merge them. I trust my computer more when it comes to this :)

ianpartridge:
Awesome. Thanks so much for taking on this extra work.

cgossain:
@ianpartridge no problem Ian, I'm very particular when it comes to these types of things! I've just submitted an update to the PR.

ianpartridge:
Thanks so much!

Issue #1450, opened 2019-05-14
Kitura listens on every available address/interface


TaborKelly:
## Context and Description
Kitura listens on every available address/interface. This is okay when you are running in a Docker container but becomes a real problem when you aren't. For example, I only want Kitura to listen on localhost so that I can force everyone to go through an nginx reverse-proxy.

### Environment Details
Ubuntu 18.04 with Swift 5.0.1.

### Steps to Reproduce
1) Use Kitura to `Kitura.addHTTPServer(onPort: port, ...); Kitura.start()`
2) Notice that Kitura is now listening on `port` for every address/interface.

### Expected vs. Actual Behaviour
I would very much like an option to only listen on one address.

djones6:
This is now available in Kitura-net `2.3.0` and Kitura-NIO tagged `2.1.0`.

The remaining piece is to update the `Kitura.addXXXXServer` API to expose the new option.

TaborKelly:
Can you re-open this issue, since the Kitura changes aren't done yet?

TaborKelly:
Would it be possible to get a tagged release with this change?

djones6:
@TaborKelly this requires a minor version increment as it adds new API.  I expect that we'll do a Kitura 2.8 release next week.

djones6:
This is now tagged in Kitura `2.8.0`.

PR #1453, opened 2019-05-24
Add option to only listen on one address

TaborKelly:
This fixes https://github.com/IBM-Swift/Kitura/issues/1450.

## Description
Add address parameter to `Kitura.addHTTPServer()` and `Kitura.addFastCGIServer()` which are passed to Kitura-Net and Kitura-NIO for `Server.listen()`.

Feedback is very welcome, especially if I accidentally missed some of your coding convention. I would pay special attention to the parameter order that I used for the methods.

## Motivation and Context
This is necessary to fix https://github.com/IBM-Swift/Kitura/issues/1450.

## How Has This Been Tested?
Manual testing by myself. I also modified the test harness to use the new code path.

## Checklist:

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -22,9 +22,9 @@ import Foundation
 var kituraNetPackage: Package.Dependency
 
 if ProcessInfo.processInfo.environment["KITURA_NIO"] != nil {
-    kituraNetPackage = .package(url: "https://github.com/IBM-Swift/Kitura-NIO.git", from: "2.0.0")
+    kituraNetPackage = .package(url: "https://github.com/IBM-Swift/Kitura-NIO.git", from: "2.1.0")
 } else {
-    kituraNetPackage = .package(url: "https://github.com/IBM-Swift/Kitura-net.git", from: "2.2.0")
+    kituraNetPackage = .package(url: "https://github.com/IBM-Swift/Kitura-net.git", from: "2.3.0")
 }
 
 let package = Package(
--- a/Sources/Kitura/Kitura.swift
+++ b/Sources/Kitura/Kitura.swift
@@ -226,12 +226,12 @@ public class Kitura {
         serverLock.lock()
         var numberOfFailures = 0
         for (server, port, address) in httpServersAndPorts {
-            Log.verbose("Starting an HTTP Server on port \(port) with address \(String(describing: address))...")
+            Log.verbose("Starting an HTTP Server on \(address ?? "*"):\(port)...")
             do {
                 try server.listen(on: port, address: address)
             } catch {
                 numberOfFailures += 1
-                Log.error("Error listening on port \(port): \(error). Use server.failed(callback:) to handle")
+                Log.error("Error listening on \(address ?? "*"):\(port): \(error). Use server.failed(callback:) to handle")
             }
         }
         for (server, path) in httpServersAndUnixSocketPaths {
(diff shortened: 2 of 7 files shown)


Package.swift
```diff
@@ -22,9 +22,9 @@ import Foundation
 var kituraNetPackage: Package.Dependency
 
 if ProcessInfo.processInfo.environment["KITURA_NIO"] != nil {
-    kituraNetPackage = .package(url: "https://github.com/IBM-Swift/Kitura-NIO.git", from: "2.0.0")
+    kituraNetPackage = .package(url: "https://github.com/IBM-Swift/Kitura-NIO.git", from: "2.1.0")
```
djones6: These changes will need applying to the `Package@swift-4.x.swift` files too.

Sources/Kitura/Kitura.swift
```diff
@@ -56,21 +56,25 @@ public class Kitura {
     ///### Usage Example: ###
     ///```swift
     /// let router = Router()
-    /// Kitura.addHTTPServer(onPort: 8080, with: router)
+    /// Kitura.addHTTPServer(onPort: 8080, with: router, address: "localhost")
     ///```
     /// - Parameter onPort: The port to listen on.
     /// - Parameter with: The `ServerDelegate` to use.
+    /// - Parameter address: The address to listen on, for example "localhost". The default is nil, which listens on
+    ///             all addresses.
```
djones6: I think it makes sense to put the address param after the port, before the delegate (router), so that related values are together.

Also I prefer the name `onAddress` as it matches with `onPort`.

Sources/Kitura/Kitura.swift
```diff
@@ -212,16 +219,16 @@ public class Kitura {
...
+    /// Kitura.addHTTPServer(onPort: 8080, with: router, address: "localhost")
     /// Kitura.startWithStatus() // Returns the number of failed server starts.
     ///```
     public class func startWithStatus() -> Int {
         serverLock.lock()
         var numberOfFailures = 0
-        for (server, port) in httpServersAndPorts {
-            Log.verbose("Starting an HTTP Server on port \(port)...")
+        for (server, port, address) in httpServersAndPorts {
+            Log.verbose("Starting an HTTP Server on port \(port) with address \(String(describing: address))...")
```
djones6: I _think_ this will result in `Optional(value)`, can I suggest using a default value (eg. '*') to avoid this.
Example in the version I mocked up here: a5539cf103ea1e3047b23d64b633160b091b4c4f

Package.swift
TaborKelly: I'll update them, but I don't understand why the Jenkins passed?

Sources/Kitura/Kitura.swift
```diff
@@ -34,15 +34,15 @@ import Dispatch
      response.send("Hello world")
      next()
  }
- Kitura.addHTTPServer(onPort: 8080, with: router)
+ Kitura.addHTTPServer(onPort: 8080, with: router, address: "localhost")
```
pushkarnk: Please update this doc comment with the right parameter names and order.

TaborKelly:
Does this look good?

Package.swift
djones6: In the absence of other constraints, SPM will choose the highest available release from the specified major, so `2.0.0` and `2.1.0` will both resolve to `2.1.0`.  However, if a user already had `2.0.x` (eg. in an existing build or a Package.resolved file) then that would satisfy the minimum version declared in the dependency, and the build would fail.  This won't occur in CI testing, as the build is always started from a clean state.

djones6:

Issue #1329, opened 2018-09-07
[StaticFileServer] Option to fall back to index.html on 404 not found


RahulKatariya:
## Context and Description
I am building an application with React and Browser Router. As it is a single page application, it requires to fall back to index.html on 404 not found.

**Note**: It works fine with the hash-based routing where URLs have `#` like `https://localhost:8080/#/index.html` but this type of routing doesn't use HTML5 history API and loses the state when navigating back and forward through the browser controls.

Related Resources: 
1. https://github.com/bripkens/connect-history-api-fallback
2. https://webpack.js.org/configuration/dev-server/#devserver-historyapifallback
3. https://developer.mozilla.org/en-US/docs/Web/API/History

### Environment Details
Information about your OS, Swift version and Xcode version if on macOS.
Swift: 4.1.2

### Steps to Reproduce
The StaticFileServer used with any front-end framework (Angular, React or Vue) and Browser router instead of Hash router will reproduce the issue.

### Expected vs. Actual Behaviour
The expected behaviour is to fall back to index.html and the application takes care of the routing while the Kitura framework responded with `Cannot GET /{PATH}`

kilnerm:
It appears the behavior requested is present on other run-times the provide file server capabilities.

We will need to add this functionality via a configuration option on the static file server so users can create file servers that exhibit this behavior.

pushkarnk:
Received a sample that reproduces the problem from Rahul. I played around and can see the problem. Thanks @RahulKatariya

pushkarnk:
Prototyped a new `StaticFileServer.Option` named `fallbackToDefaultIndex` 
https://github.com/pushkarnk/Kitura/commit/718d6a21e528d0c603a053f104604b0dfa1787d9

Got the running webapp reviewed from Rahul. The expectation is a little different from what I had assumed: on the user trying to bypass the landing page and accessing a location directly, the default landing location (like `/index.html`) must be rendered, but the user entered path must be retained. For example, if the user tries to access `/user/report`, the index.html pertaining to  `localhost:8080` must be rendered but the path must be retained as `localhost:8080/user/report`

pushkarnk:
The feedback above means a few more changes to the prototype. 

Apart from that, there are a few things to take care of here(just thinking aloud):
1. The user must be able to configure the fallback location. The default could be `/index.html`.
2. No fallback if the user is trying to directly access a file
3. The user must be able to configure locations for which the fallback fails

[...]

PR #1464, opened 2019-06-14
Add a StaticFileServer option to fallback to the index page

pushkarnk:
Add a StaticFileServer option to fallback to the index page

## Description
This commit adds a new option to the `StaticFileServer.Options` class to enable
fallback to the index.html. This is intended to be used only by single page
applications. Just before the static file server returns a 404, we check if
this option is configured and serve the default index page if it is. The webapp
takes care of the rendering.

## Motivation and Context
Single page applications use only one index file that is accessible by web
browsers, which is usually the index.html, which also serves as a landing page.
All the other pages are accessible only from the landing page. If the users
bypasses the index page to directly visit an inner page, or if they try to
refresh an inner page, the static file server returns a 404  error. Some SPAs
might wish to have a fallback feature where the static file server
serves the index.html in the event of such user activity.

This is in response to #1329 

## How Has This Been Tested?
This has been tested with a user's single-page application.

## Checklist:

## Caveat
In a single page application, when we decide to serve the index page when anything other than the landing page is requested for, we fall into a tricky situation when it comes to responding to invalid paths. For example, `/help/contact` may be a valid path but `/help/contact/dfdfdf` may be in invalid path. With the fallback option in place, it is difficult to differentiate between a valid non-landing page path and an invalid page. There are two different approaches followed by some popular single-page apps:
 - Default to index: example: `mail.google.com` accepts any invalid path like `https://mail.google.com/mail/#inbox/dsdsd/dsds` and falls back to `https://mail.google.com/mail/#inbox`, returning HTTP status code 200.
 -  Return 404: example: `twitter.com` returns 404 for invalid paths.

With the current implementation, we'd always return status code 200. The javascript running in the browser can detect an invalid path and still show a message indicating 404.

### DIFF
--- a/Sources/Kitura/staticFileServer/FileServer.swift
+++ b/Sources/Kitura/staticFileServer/FileServer.swift
@@ -41,6 +41,8 @@ extension StaticFileServer {
         /// Whether accepts range requests or not
         let acceptRanges: Bool
 
+        /// A default index to be served if the requested path is not found.
+        /// This is intended to be used by single page applications.
         let defaultIndex: String?
 
         init(servingFilesPath: String, options: StaticFileServer.Options,
@@ -108,6 +110,11 @@ extension StaticFileServer {
                 return
             }
 
+            // We haven't been able to find the requested path. For single page applications,
+            // a default fallback index may be configured. Before we serve the default index
+            // we make sure that this is a not a direct file request. The best check we could
+            // run for that is if the requested file contains a '.'. This is inspired by:
+            // https://github.com/bripkens/connect-history-api-fallback
             let isDirectFileAccess = filePath.split(separator: "/").last?.contains(".") ?? false
             if isDirectFileAccess == false, let defaultIndex = self.defaultIndex {
                 serveDefaultIndex(defaultIndex: defaultIndex, response: response)
--- a/Sources/Kitura/staticFileServer/FileServer.swift
+++ b/Sources/Kitura/staticFileServer/FileServer.swift
@@ -41,6 +41,8 @@ extension StaticFileServer {
         /// Whether accepts range requests or not
         let acceptRanges: Bool
 
+        let fallbackToDefaultIndex: Bool
+
         init(servingFilesPath: String, options: StaticFileServer.Options,
              responseHeadersSetter: ResponseHeadersSetter?) {
             self.possibleExtensions = options.possibleExtensions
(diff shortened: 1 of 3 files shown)


RahulKatariya:
Hey @pushkarnk, 

On the Caveats, In a single page application, only the resources are valid and all other paths should default to index.html. As we default to index.html and we are also retaining the url in the browser, Apps can takes care of showing 404, if the path in the url is invalid.

In your example where /help/contact may be a valid path but /help/contact/dfdfdf may be in invalid path, Both of the paths are not resources so they are redirected to index.html. The index of the app has the logic to parse the path, and it's the developer choice whether to show 404 on `/help/contact/dfdfdf` or redirect to `/help/contact/`.

pushkarnk:
@RahulKatariya Thanks for the explanation!

Sources/Kitura/staticFileServer/FileServer.swift
```diff
@@ -41,6 +41,8 @@ extension StaticFileServer {
         /// Whether accepts range requests or not
         let acceptRanges: Bool
 
+        let defaultIndex: String?
```
djones6: Can we add a comment describing this property?

Sources/Kitura/staticFileServer/FileServer.swift
```diff
@@ -101,14 +104,31 @@ extension StaticFileServer {
                 return
             }
 
-            tryToServeWithExtensions(filePath, response: response)
+            if tryToServeWithExtensions(filePath, response: response) {
+                return
+            }
+
+            let isDirectFileAccess = filePath.split(separator: "/").last?.contains(".") ?? false
```
djones6: Could we add a comment referencing the logic from Express that we are replicating here? (for future reference)

Sources/Kitura/staticFileServer/StaticFileServer.swift
```diff
@@ -77,13 +78,24 @@ open class StaticFileServer: RouterMiddleware {
         /// - Parameter redirect: an indication whether to redirect to trailing
         /// "/" when the requested path is a directory.
         /// - Parameter cacheOptions: cache options for StaticFileServer.
+        /// - Parameter defaultIndex: A default index, like "/index.html", to be served if the
+        /// requested path is not found. This is intended to be used by single page applications
+        /// that wish to fallback to a default index when a requested path is not found.
```
djones6: Given the addition of the check for a file request, is it worth adding something like this?
>when a requested path is not found, and where that path is not a file request.

djones6:

PR #1466, opened 2019-06-19
Swift 5.1 support

djones6:
## Description

Resolve compilation errors and warnings relating to changes in the Swift 5.1 release:
- `import FoundationNetworking` on Linux to access `HTTPCookie` and similar types, due to [changes to Foundation structure](https://forums.swift.org/t/foundationnetworking/24769)
- fix compilation warnings relating to `var`s that were never mutated
- Temporarily disables two of the TestCodableRouter tests that have extremely long compile times with Swift 5.1 on Linux (reported under: https://bugs.swift.org/browse/SR-11012)

Depends on https://github.com/IBM-Swift/Package-Builder/pull/169 to deploy workaround for running 5.1 snapshot in Docker on Travis.

## Checklist:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ matrix:
       dist: xenial
       sudo: required
       services: docker
-      env: DOCKER_IMAGE=swift:5.0.1 SWIFT_SNAPSHOT=$SWIFT_DEVELOPMENT_SNAPSHOT SWIFT_TEST_ARGS="--parallel"
+      env: SWIFT_SNAPSHOT=$SWIFT_DEVELOPMENT_SNAPSHOT SWIFT_TEST_ARGS="--parallel"
     - os: osx
       osx_image: xcode9.2
       sudo: required
@@ -72,10 +72,11 @@ matrix:
       osx_image: xcode10.2
       sudo: required
       env: KITURA_NIO=1 SWIFT_TEST_ARGS="--parallel"
-    - os: osx
-      osx_image: xcode10.2
-      sudo: required
-      env: SWIFT_SNAPSHOT=$SWIFT_DEVELOPMENT_SNAPSHOT SWIFT_TEST_ARGS="--parallel"
+# Awaiting xcode11 image from Travis
+#    - os: osx
+#      osx_image: xcode10.2
+#      sudo: required
+#      env: SWIFT_SNAPSHOT=$SWIFT_DEVELOPMENT_SNAPSHOT SWIFT_TEST_ARGS="--parallel"
 
 before_install:
   - git clone https://github.com/IBM-Swift/Package-Builder.git
--- a/.travis.yml
+++ b/.travis.yml
@@ -72,7 +72,7 @@ matrix:
       osx_image: xcode10.2
       sudo: required
       env: KITURA_NIO=1 SWIFT_TEST_ARGS="--parallel"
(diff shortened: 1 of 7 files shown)


djones6:
For some reason, trying to install and use the 5.1 toolchain inside the swift:5.0.1 Docker image is failing (https://travis-ci.org/IBM-Swift/Kitura/jobs/547756128) with:
`error: toolchain is invalid: could not find the `swiftc` at expected path /home/travis/build/IBM-Swift/Kitura/swift-5.1-DEVELOPMENT-SNAPSHOT-2019-06-16-a-ubuntu18.04/usr/bin/swiftc`
... however I cannot recreate this failure outside of Travis (even running the exact same commands that Travis is running, ie. to run the build via Package-Builder inside a swift:5.0.1 container).  It just works.

The `swiftc` command is a symlink to `swift` in the same directory, and Package-Builder unpacks the swift toolchain to the project's directory.  When running under Docker, this directory is a mounted volume.  I wonder whether there's some property of the filesystem that Travis provides that, when mounted inside Docker as a volume, causes this symlink not to work?

In ba897be I tried taking the Docker layer out (running the same 5.1 toolchain directly in Travis' 16.04 image).  This works in that we successfully build Kitura using that toolchain, but it then seems to hang while building the the `TestTypeSafeMiddleware` test (https://travis-ci.org/IBM-Swift/Kitura/jobs/547790233).   However, I _do_ seem to be able to recreate this locally (on 16.04).  Doesn't hang on 18.04.

djones6:
Update: It isn't a hang, just takes several minutes to compile.  And I _do_ see the same behaviour on 18.04, it's not 16.04-specific, I must have misled myself somehow.

Seems the output from swift build is a little misleading.  Although it usually pauses when it says `[mm/nn] Compiling KituraTests TestTypeSafeMiddleware.swift`, after a process of elimination, it's actually `TestCodableRouter.swift` that is taking a long time to compile.

That source file contains a large number of tests, so I'm in the process of narrowing down which test(s) specifically are causing the problem.  Once done, we can guard the offending ones with a version check and raise a bug.

djones6:
Finally got to the bottom of the failures of 5.1 toolchains under Docker running in Travis.
It relates to https://github.com/apple/swift-corelibs-foundation/pull/1837 which introduces the use of the `statx` syscall to obtain file attributes.  This is only available in 'recent' Linux kernels (4.11, circa April 2017). and Foundation checks that it is available.

However, even if it is available, we may not have permission to call it.  It requires Docker `18.04` and libseccomp `2.3.3` or higher.  Both are available on 16.04 and 18.04.

Travis' 16.04 (xenial) environment provides `libseccomp2 2.3.1-2.1ubuntu2~16.04.1`, which is too old.  As a result, even though Docker is recent enough, the `statx` call is not permitted, and so Foundation throws an error, which is [swallowed by SPM](https://github.com/apple/swift-package-manager/blob/swift-5.1-branch/Sources/Basic/FileSystem.swift#L250) - and only [surfaced eventually as a "swiftc not found"](https://github.com/apple/swift-package-manager/blob/swift-5.1-branch/Sources/Workspace/UserToolchain.swift#L96) message.

The fix for this is to update Foundation so that if the statx call is available, but results in an `errno` of `1` (EPERM - not permitted), it falls back to the earlier `stat` approach.

A possible workaround is to upgrade `libseccomp2` before running under Docker in Travis.

PR #1468, opened 2019-06-24
A convenience API to create and add cookies.

### LINKED ISSUE #1456: Cookie creation enhancement
## Context and Description
Kitura uses Foundations `HTTPCookie`, which is great. However, it would be nice if it were easier to create `HTTPCookie`s. 

### Environment Details
Swift 5.0.1, XCode 10.2.1, macOS 10.14.5.

### Steps to Reproduce
Try to create a cookie to send in a `RouterResponse`. You end up writing something like this:
```swift
func createCookie(key: String, value: String, hostname: String) -> HTTPCookie? {
    var cookieProps: [HTTPCookiePropertyKey: Any]
    cookieProps = [
        HTTPCookiePropertyKey.domain: hostname,
        HTTPCookiePropertyKey.path: "/",
        HTTPCookiePropertyKey.name: key,
        HTTPCookiePropertyKey.value: value
    ]

    let cookie = HTTPCookie(properties: cookieProps)
    return cookie
}
```

It would be great to have a helper function for this in Kitura.

RudraniW:
A convenience API to create and add cookies.

## Description
The new public enum `CookieAttribute` is created with cases that represent the attributes of a cookie. The `addCookie` function takes an array of  type `CookieAttribute` and returns either an HTTPCookie object or nil, depending on whether foundation HTTPCookie initializer passes or fails. Each attribute in the array is added into the dictionary `cookieProperties` with appropriate keys. The attributes having nil values are not added. This `cookieProperties` dictionary is then passed to the HTTPCookie initializer. If this initializer returns a new  HTTPcookie object, with the given properties, the HTTPCookie object is added to the `cookies` dictionary which contains set of cookies to return with the response. If the initializer fails then a warning is logged. Resolves #1456

## Motivation and Context
Currently, the user has to write a lot of boilerplate code for cookie creation. This pull request tries to cookie creation and its addition to the response easier for the user.

## Checklist:

### DIFF
--- a/Sources/Kitura/OptionalCookieAttribute.swift
+++ b/Sources/Kitura/OptionalCookieAttribute.swift
@@ -1,7 +1,7 @@
 import Foundation
 
 /// Describes an optional attribute of a cookie.
-public struct OptionalCookieAttribute {
+public struct AdditionalCookieAttribute {
     internal enum _CookieAttribute {
         // A comment for the cookie.
         case comment(String?)
@@ -34,54 +34,54 @@ public struct OptionalCookieAttribute {
     }
 
     /// A comment for the cookie.
-    public static func comment(_ value: String?) -> OptionalCookieAttribute {
-        return OptionalCookieAttribute(_CookieAttribute.comment(value))
+    public static func comment(_ value: String?) -> AdditionalCookieAttribute {
+        return AdditionalCookieAttribute(_CookieAttribute.comment(value))
     }
 
     /// A URL that can be presented to the user as a link for further information about this cookie.
-    public static func commentURL(_ value: String?) -> OptionalCookieAttribute {
-        return OptionalCookieAttribute(_CookieAttribute.commentURL(value))
+    public static func commentURL(_ value: String?) -> AdditionalCookieAttribute {
+        return AdditionalCookieAttribute(_CookieAttribute.commentURL(value))
     }
 
     /// Custom cookie attributes
     ///
     ///Note: Custom cookie attributes are not honoured by Foundation yet.
-    public static func custom(_ key: String, _ value: String) -> OptionalCookieAttribute {
-        return OptionalCookieAttribute(_CookieAttribute.custom(key, value))
+    public static func custom(_ key: String, _ value: String) -> AdditionalCookieAttribute {
+        return AdditionalCookieAttribute(_CookieAttribute.custom(key, value))
     }
 
     /// A String value representing a boolean (TRUE/FALSE), stating whether the cookie should be discarded at the end of the session.
-    public static func discard(_ value: String) -> OptionalCookieAttribute {
-        return OptionalCookieAttribute(_CookieAttribute.discard(value))
+    public static func discard(_ value: String) -> AdditionalCookieAttribute {
+        return AdditionalCookieAttribute(_CookieAttribute.discard(value))
     }
 
     /// The cookie’s expiration date. The expiration date is the date when the cookie should be deleted.
-    public static func expires(_ value: Date?) -> OptionalCookieAttribute {
-        return OptionalCookieAttribute(_CookieAttribute.expires(value))
+    public static func expires(_ value: Date?) -> AdditionalCookieAttribute {
+        return AdditionalCookieAttribute(_CookieAttribute.expires(value))
     }
 
     /// A boolean value that indicates whether this cookie should only be sent over secure channels.
-    public static func isSecure(_ value: Bool) -> OptionalCookieAttribute {
-        return OptionalCookieAttribute(_CookieAttribute.isSecure(value))
+    public static func isSecure(_ value: Bool) -> AdditionalCookieAttribute {
+        return AdditionalCookieAttribute(_CookieAttribute.isSecure(value))
     }
 
     /// A value stating how long in seconds the cookie should be kept, at most.
-    public static func maximumAge(_ value: String) -> OptionalCookieAttribute {
-        return OptionalCookieAttribute(_CookieAttribute.maximumAge(value))
+    public static func maximumAge(_ value: String) -> AdditionalCookieAttribute {
+        return AdditionalCookieAttribute(_CookieAttribute.maximumAge(value))
     }
 
     /// The URL that set this cookie.
-    public static func originURL(_ value: URL?) -> OptionalCookieAttribute {
-        return OptionalCookieAttribute(_CookieAttribute.originURL(value))
+    public static func originURL(_ value: URL?) -> AdditionalCookieAttribute {
+        return AdditionalCookieAttribute(_CookieAttribute.originURL(value))
     }
 
     /// The list of ports for the cookie,  an array of NSNumber objects containing integers.
-    public static func portList(_ value: [NSNumber]?) -> OptionalCookieAttribute {
-        return OptionalCookieAttribute(_CookieAttribute.portList(value))
+    public static func portList(_ value: [NSNumber]?) -> AdditionalCookieAttribute {
+        return AdditionalCookieAttribute(_CookieAttribute.portList(value))
     }
 
     /// The version of the cookie. Must be either 0 or 1. The default is 0.
-    public static func version(_ value: Int) -> OptionalCookieAttribute {
-        return OptionalCookieAttribute(_CookieAttribute.version(value))
+    public static func version(_ value: Int) -> AdditionalCookieAttribute {
+        return AdditionalCookieAttribute(_CookieAttribute.version(value))
     }
 }
--- a/Sources/Kitura/OptionalCookieAttribute.swift
+++ b/Sources/Kitura/OptionalCookieAttribute.swift
@@ -1,0 +1,87 @@
+import Foundation
+
+/// Describes an optional attribute of a cookie.
+public struct OptionalCookieAttribute {
+    internal enum _CookieAttribute {
+        // A comment for the cookie.
+        case comment(String?)
+        // A URL that can be presented to the user as a link for further information about this cookie.
+        case commentURL(String?)
+        // Custom cookie attributes
+        case custom(String, String)
+        // A String value representing a boolean (TRUE/FALSE), stating whether the cookie should be discarded at the end of the session
+        case discard(String)
+        // The list of ports for the cookie,  an array of NSNumber objects containing integers.
+        case expires(Date?)
+        //  A boolean value that indicates whether this cookie should only be sent over secure channels.
+        case isSecure(Bool)
+        // A value stating how long in seconds the cookie should be kept, at most.
+        case maximumAge(String)
+        // The URL that set this cookie.
+        case originURL(URL?)
+        // The version of the cookie. Must be either 0 or 1. The default is 0.
+        case portList([NSNumber]?)
+        // The cookie’s expiration date. The expiration date is the date when the cookie should be deleted.
+        case version(Int)
+    }
+
+    // The internal case represented by this instance of CookieAttribute.
+    internal let _value: _CookieAttribute
+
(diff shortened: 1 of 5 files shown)


Sources/Kitura/RouterResponse.swift
```diff
@@ -174,6 +174,116 @@ public class RouterResponse {
...
+        /// A comment for the cookie.
+        case comment(String?)
+
+        /// A URL that can be presented to the user as a link for further information about this cookie.
+        case commentURL(String?)
+    }
+
+    /// Add a cookie to the response.
+    ///
+    /// This function attempts to create an `HTTPCookie`  from a given `CookieAttribute`  array and, if successful , adds the HTTPCookie to the `cookies` dictionary.
```
TaborKelly: You should probably mention that for this to succeed the caller will need to provide (at a minimum) the `path`, `name`, and `value` attributes, and either the `originURL` or `domain` [as per the Apple documentation](https://developer.apple.com/documentation/foundation/httpcookie/1392975-init).

RudraniW:
@djones6, @kilnerm 
Can you please review this change?
This PR is the solution for cookie creation enhancement request: https://github.com/IBM-Swift/Kitura/issues/1456. With this new API if user doesn't give mandatory attributes name, value, domain and path, Foundation's HTTPCookie initializer fails. To avoid this problem I have implemented an alternative solution, where mandatory attributes are function parameters. 
You can check the alternate solution here : https://github.com/RudraniW/Kitura/commit/1dd52f3ea06375c8e8ecb3f56629098a31ea09c0

Sources/Kitura/RouterResponse.swift
```diff
@@ -178,6 +178,116 @@ public class RouterResponse {
             }
         }
     }
+    
+    /// An enum to describe the attributes  (name, value, domain, path etc) of a cookie.
+    public enum CookieAttribute {
```
kilnerm: It looks like HTTPCookiePropertyKey allows you to define custom cookie property keys, can we add support for that behavior?

Is it possible that the available HTTPCookiePropertyKey value set changes in the future? If so then maybe an alternative to enum would be a good idea so we avoid the need for a major version due to changing a public enum.

Sources/Kitura/RouterResponse.swift
```diff
@@ -178,6 +178,116 @@ public class RouterResponse {
...
+        case commentURL(String?)
+    }
+
+    /// Add a cookie to the response.
+    ///
+    /// This function attempts to create an `HTTPCookie`  from a given `CookieAttribute`  array and, if successful , adds the HTTPCookie to the `cookies` dictionary.
+    /// - Parameter with: An  array of  `CookieAttribute`s
+    /// - Returns: Returns the added `HTTPCookie` or `nil` in the case of  an `HTTPCookie` creation failure. Here success or failure is governed by the Foundation's HTTPCookie specifiacation: https://developer.apple.com/documentation/foundation/httpcookie
+    @discardableResult
+    public func addCookie(with attributes: [CookieAttribute]) -> HTTPCookie? {
```
kilnerm: As mentioned by TaborKelly, if some properties are required it would be good to separate them from the optional properties. By doing so we are making it easier for the user to write correct code and aligning better with the language principles.

Sources/Kitura/RouterResponse.swift
RudraniW: Thanks for the review .I will find a way to add support for defining custom cookie property keys. I also work on keeping the enum extensible.

Sources/Kitura/RouterResponse.swift
RudraniW: Can you please review this solution?
I have implemented `addCookie` with mandatory attributes name, value, domain and path as function parameters. 
The link to the implementation is : https://github.com/RudraniW/Kitura/commit/1dd52f3ea06375c8e8ecb3f56629098a31ea09c0

RudraniW:
@kilnerm I have addressed your review comments. 
- I have added support for custom cookie attributes. But these custom cookie attributes are not honored by Foundation on both Linux and Darwin.
- I have implemented `addCookie` with mandatory cookie attributes name, value, domain and path as function parameters so that HTTPCookie initializer never fails.
- I considered implementing `CookieAttribute` enum in open enum style but it is making code too much complex because different cases have different associated types. Also it is unlikely that Foundation will add new cookie attributes to HTTPCookiePropertyKey.

Sources/Kitura/RouterResponse.swift
kilnerm: Thanks. Given the API exists it is probably better to expose it, particularly as we are sticking with an enum!

djones6: I'm not convinced that we should be introducing a public enum as a proxy for something that is not under our control (even though Foundation is unlikely to change), it leaves us exposed if we were forced to change it for some reason.

You could instead create a Struct with the same semantics as the enum, using an internal enum (so we can change it in the future without breaking our public API).  It looks something like this:

```swift
public struct CookieAttribute {
    internal enum _CookieAttribute {
        case portList([NSNumber]?)
        ... etc
    }
    internal let _value: CookieAttribute

    private init(_ value: _CookieAttribute) {
        self._value = value
    }

    public static func portList(_ value: [NSNumber]?) -> CookieAttribute {
        return CookieAttribute(_CookieAttribute(value))
    }
    ...etc
}
```

djones6:
@RudraniW Here's my suggestion hacked on top of your PR: https://github.com/djones6/Kitura/commit/61651da4bbcb5095438d92b3414d829ca1e00dcb
(you're welcome to use my commit if you like, or use it as the basis for your own changes)
@kilnerm @pushkarnk what do you think about this approach instead?

pushkarnk:
@djones6 Sure, we can have an open `CookieAttribute` enum as you've demonstrated. Rudrani and I discussed using `static functions` for associated values but we weren't clear about how we'll do the deconstruction/parsing. 

However, I'm a little uncomfortable about not having the mandatory cookie attributes included in the `CookieAttributes`. Given that it does not include all the cookie attributes, what would we call it?

djones6:
I think it's going to cause surprise / confusion for somebody, if you include name/path/etc in the set (but then tell people not to use them).

Personally I'd think it's still fine to call it `CookieAttributes`, but I suppose it could be called `OptionalCookieAttributes`?

RudraniW:
@djones6 Thanks for the feedback. I have addressed your review comments.
- I have implemented `CookieAttribute` enum in open enum style.
- I have removed the mandatory attributes name, value, path and domain from the enum.

djones6:
The code is failing the swift linter:
```
/Users/travis/build/IBM-Swift/Kitura/Sources/Kitura/RouterResponse.swift:184:18: error: Type Name Violation: Type name should only contain alphanumeric characters: '_CookieAttribute' (type_name)
```
Not sure if we want to squash this linter error or rename? @ianpartridge what's your opinion?

pushkarnk:
I'd vote for removing this linter rule. It's common practice to have internal type names starting with `_`.

Sources/Kitura/RouterResponse.swift
```diff
@@ -179,6 +179,143 @@ public class RouterResponse {
         }
     }
 
+    /// Describes an optional attribute of a cookie.
+    public struct CookieAttribute {
```
ianpartridge: I'm sorry, I don't see why `struct CookieAttribute` belongs as nested inside `RouterResponse`. Can you explain the rationale for that?

ianpartridge:
Turn off the linter error - a `_` prefix is a Swift-ism.

Sources/Kitura/RouterResponse.swift
RudraniW: As cookies are closely related with HTTP response, we thought of adding `struct CookieAttribute` inside `RouterResponse`. If you think we should keep this strust independent, I can do that.

Sources/Kitura/RouterResponse.swift
```diff
@@ -179,6 +179,143 @@ public class RouterResponse {
...
+
+    /// Add a cookie to the response.
+    ///
+    /// This function creates an `HTTPCookie`  from the provided attributes and adds it to the `cookies` dictionary.
+    /// - Parameter name: The cookie’s name.
+    /// - Parameter value: The cookie‘s value.
+    /// - Parameter domain: The domain of the cookie.
+    /// - Parameter path: The cookie’s path.
+    /// - Parameter otherAttributes: An array of  any other optional cookie attributes
+    public func addCookie(name: String, value: String, domain: String, path: String, otherAttributes: [CookieAttribute] = []) {
```
ianpartridge: This will allocate an array for `otherAttributes` every time, even if it's not used and is empty. It would be better to make it an optional parameter with a default of `nil`.

Sources/Kitura/RouterResponse.swift
ianpartridge: @djones6 @pushkarnk what do you think?

pushkarnk: Hmm. Though I was involved in the decision of making this a nested type,  I'm now thinking of how the name of the type sounds. `RouterResponse.CookieAttribute` doesn't sound logical ( like `RouterResponse.Cookie` or `Cookie.CookieAttribute` would have).

djones6: I agree, though I suspect inventing a top level type called 'Cookie' to nest it in would be asking for trouble.

Could we make a public extension to HTTPCookie or is that bad practice?

ianpartridge: Bad practice I think.

djones6:
I see that the struct was moved to be a top-level `CookieAttribute` type, and its internal enum renamed to `OptionalCookieAttribute`.  I guess this gets around the linter problem, but I'd argue the names should be the other way around:  `OptionalCookieAttribute` is more descriptive, so should be the public type name.  I'd still prefer the internal enum be named the same (with `_` prefix), as I think the intent is clearer.

If the new type is going to be a top-level type, I'd prefer it to be defined in its own source file (ie. `OptionalCookieAttribute.swift`).

@RudraniW To disable the linter rule, add `type_name` to the disabled rules in `.swiftlint.yml`

RudraniW:
@djones6 Thanks for the feedback. 
I have addressed your review comments. I have made following changes
- Created new swift file with name `OptionalCookieAttribute.swift` which contains  public struct renamed as `OptionalCookieAttribute` and enum `_CookieAttribute`.
- Disabled the linter rule for `type_name`.

Sources/Kitura/RouterResponse.swift
```diff
@@ -179,8 +179,71 @@ public class RouterResponse {
...
+                    if let comment = comment {
+                        cookieProperties[HTTPCookiePropertyKey.comment] = comment
+                    }
+
+                case .commentURL(let commentURL):
+                    if let commentURL = commentURL {
+                        cookieProperties[HTTPCookiePropertyKey.commentURL] = commentURL
+                    }
+
+                case.custom(let key, let value):
```
djones6: Now that we're free to extend the struct at any time, do we need a `.custom`?  If not, it would be cleaner to remove it.
Otherwise, can you add the missing space between `case` and `.custom` here.

Sources/Kitura/OptionalCookieAttribute.swift
```diff
@@ -0,0 +1,87 @@
+import Foundation
+
+/// Describes an optional attribute of a cookie.
+public struct OptionalCookieAttribute {
```
djones6: Having talked naming with @ianpartridge, we prefer the name `AdditionalCookieAttribute`, to avoid confusion with Swift's `Optional` type.  Please could you rename the type (and the source file)?

PR #1481, opened 2019-08-16
feat: Ability to limit request size and connection count

### LINKED ISSUE #1384: Ability to limit incoming request size
## Context and Description
Kitura currently does not have the concept of a limit on incoming request size.  It is possible to destabilize a server by sending (accidentally or maliciously) a large request payload which either impairs performance, or exhausts available memory.

A limit should be configurable on a server which causes requests above that size to be rejected, and the connection closed.

djones6:
## Description

This PR allows a `requestSizeLimit` (maximum bytes for the request body) and `connectionLimit` (on total number of concurrent connections) to be configured when registering a server.

Example usage from Kitura is as follows:
```swift
let options = ServerOptions(requestSizeLimit: 100, connectionLimit: 3)
Kitura.addHTTPServer(onPort: port, with: router, options: options)
```

If you do not specify either of the parameters, or do not specify a `ServerOptions` at all, then default values are used.  The defaults are defined as constants `ServerOptions.defaultRequestSizeLimit` and `ServerOptions.defaultConnectionLimit`.
- These have been given conservative defaults of 100mb and 10,000 connections, so as to not break existing deployments.

If a client sends a request that exceeds the limit, then they are immediately sent an `HTTP/1.1 413 Request Entity Too Large` and the connection is closed.
If a client attempts to connect, but we are already at the connection limit, they are immediately sent an `HTTP/1.1 503 Service Unavailable` and the connection is closed.

These responses can be further customized by supplying `requestSizeResponseGenerator` and `connectionResponseGenerator` closures, which are passed an `Int` (the limit that was exceeded) and a `String` (the hostname and port of the socket), and return an `HTTPStatusCode` and a `String` that is a plaintext message to be returned to the client.  The default responses are again defined as constants `defaultRequestSizeResponseGenerator` and `defaultConnectionResponseGenerator`.

This does not provide a means to set a limit on the request _headers_ (more specifically, all data up to the point where the HTTP parser determines that headers are complete).  Both the HTTPParser in Kitura-net and Swift-NIO set a limit of 80kb on the size of the headers, and it seems unlikely in practice that this would need to be modified.

Some issues I'm aware of:
- There are already parameters `keepAlive` and `allowPortReuse` on `Kitura.addHTTPServer` which could be rolled into this new options struct.

Depends on https://github.com/IBM-Swift/Kitura-net/pull/307 and https://github.com/IBM-Swift/Kitura-NIO/pull/221

## Motivation and Context

Resolves #1384 

## How Has This Been Tested?

Tests have been added for:
- a POST request that is (exactly) within the limit is successful,
- an oversized POST request is rejected,
- a request with oversized headers is rejected,
- connections above the configured limit are rejected,
- custom responses for oversized requests and excess connections.

## Checklist:

### DIFF
--- a/Sources/Kitura/Kitura.swift
+++ b/Sources/Kitura/Kitura.swift
@@ -316,7 +316,7 @@ public class Kitura {
 
  ServerOptions allows customization of default connection policies, including:
 
- - `requestSizeLimit`: Defines the maximum size of an incoming request, in bytes. If requests are received that are larger than this limit, they will be rejected and the connection will be closed. A value of `nil` means no limit.
+ - `requestSizeLimit`: Defines the maximum size of an incoming request body, in bytes. If requests are received that are larger than this limit, they will be rejected and the connection will be closed. A value of `nil` means no limit.
  - `connectionLimit`: Defines the maximum number of concurrent connections that a server should accept. Clients attempting to connect when this limit has been reached will be rejected. A value of `nil` means no limit.
 
  The server can optionally respond to the client with a message in either of these cases. This message can be customized by defining `requestSizeResponseGenerator` and `connectionResponseGenerator`.
--- a/Sources/Kitura/Kitura.swift
+++ b/Sources/Kitura/Kitura.swift
@@ -65,7 +65,7 @@ public class Kitura {
     /// - Parameter withSSL: The `sslConfig` to use.
     /// - Parameter keepAlive: The maximum number of additional requests to permit per Keep-Alive connection. Defaults to `.unlimited`. If set to `.disabled`, Keep-Alive will not be permitted.
     /// - Parameter allowPortReuse: Determines whether the listener port may be shared with other Kitura instances (`SO_REUSEPORT`). Defaults to `false`. If the specified port is already in use by another listener that has not allowed sharing, the server will fail to start.
-    /// - Parameter connectionPolicy: Controls whether connections and requests are accepted.
+    /// - Parameter options: Allows customization of default policies for this server.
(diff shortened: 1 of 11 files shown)


Sources/Kitura/Kitura.swift
```diff
@@ -300,3 +308,30 @@ public class Kitura {
     internal private(set) static var httpServersAndUnixSocketPaths = [(server: HTTPServer, socketPath: String)]()
     internal private(set) static var fastCGIServersAndPorts = [(server: FastCGIServer, port: Port, address: String?)]()
 }
+
+// MARK: ServerOptions
+
```
djones6: Update usage example

djones6: Also need to test that this does not interfere with WebSockets / upgraded connections.

Tests/KituraTests/TestCodableRouter.swift
```diff
@@ -663,7 +663,7 @@ final class TestCodableRouter: KituraTest, KituraTestSuite {
     }
 
     func testCodableGetSingleQueryParameters() {
-        let date: Date = Coder().dateFormatter.date(from: Coder().dateFormatter.string(from: Date()))!
+        let date: Date = Coder.defaultDateFormatter.date(from: Coder.defaultDateFormatter.string(from: Date()))!
```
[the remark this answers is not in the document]
djones6: It is.  It should really be in a separate PR but it's only to fix some compilation warnings in the tests (due to https://github.com/IBM-Swift/KituraContracts/commit/5211127a5a8f92eefd199554d0ff2a1c8642381d)

Issue #1482, opened 2019-08-29
The `SwaggerDocument` struct is private and hardcoded


mbarnach:
## Context and Description
When creating a router, a private `SwaggerDocument` is created. This document is not accessible to the end user and has some hardcoded values, like `version`, `title` and `description`.
These information are used when exporting through the Kitura-OpenAPI package.
It would be handy to allow to set these values when creating the router or later.

One solution would be to set the var `swagger` has public, and well as its fields (`version`, `title` and `description`) or kept the fields private, but add them to the constructor.

PR #1483, opened 2019-08-29
Move to public the `SwaggerDocument` object of the router.

mbarnach:
Add `title`, `description` and `version` to the constructer in order to be specified by the user.
Add a related test.

## Description
The `swagger` object inside `Router` is now public.
The constructor of `SwaggerDocument` has been extended to have the `title`, `description` and `version` fields with the previous values as default.

## Motivation and Context
When generating the Swagger document, the meta data like title, description and version couldn't be changed.

Fix issue #1482.

## How Has This Been Tested?
A new test has been added.
It won't affect existing code, as it is a purely adding feature.

## Checklist:


Sources/Kitura/Router.swift
```diff
@@ -120,7 +120,7 @@ public class Router {
     // MARK: Swagger
     
     /// Contains the structures needed for swagger document generation
-    var swagger: SwaggerDocument
+    public var swagger: SwaggerDocument
```
djones6: The problem with this approach is that it introduces an ordering problem.  Routes are added to the swagger document each time you call `router.get()` et. al, and so if you were to reassign `router.swagger` after registering some routes, those routes would no longer appear in the document.

Perhaps you could leave this internal, and instead add an additional parameter eg. `apiDocument: SwaggerDocument = SwaggerDocument()` to `router.init`?

Sources/Kitura/Router.swift
```diff
@@ -92,8 +92,8 @@ public class Router {
     ///                                FileResourceServer to serve the default
     ///                                "Welcome to Kitura" page and related
     ///                                assets.
-    public init(mergeParameters: Bool = false, enableWelcomePage: Bool = true) {
-        self.swagger = SwaggerDocument()
+    public init(mergeParameters: Bool = false, enableWelcomePage: Bool = true, apiDocument: SwaggerDocument = SwaggerDocument()) {
```
djones6: This looks good, can you add the documentation comment for the new `apiDocument` parameter? Something like
> - Parameter apiDocument: Optional parameter that allows customization of the OpenAPI document describing this Router.

Sources/Kitura/Router.swift
```diff
@@ -120,7 +120,7 @@ public class Router {
     // MARK: Swagger
     
     /// Contains the structures needed for swagger document generation
-    var swagger: SwaggerDocument
+    internal var swagger: SwaggerDocument
```
djones6: `internal` is implicit, you can remove it.

mbarnach:
I think the failure in Travis is not related to the PR. Can we re-run the tests?

PR #1488, opened 2019-09-30
Add Nested Types support to Swagger Generator 

123FLO321:
## Description
This MR adds support for Nested Types to the Swagger Generator.

## Motivation and Context
This MR fixes a problem where the Swagger Generator would only use the Types name rather then the full nested path.

For example `Pokemon` and `Raid.Pokemon` would have just created `Pokemon`.

With this change the full nested path minus the first component are used when generating. 

## How Has This Been Tested?
This has been tested locally and I didn't find any problems. However more testing is likely needed to confirm that this is fully working. 

Example

## Checklist:


Sources/Kitura/SwaggerGenerator.swift
```diff
@@ -1076,6 +1075,42 @@ public struct SwaggerDocument: Encodable {
...
+    
+    private static func getTypeName(typeInfo: TypeInfo) -> String {
+        let fullName = String(reflecting: typeInfo)
+        return getTypeName(fullName: fullName)
+    }
+    
+    private static func getTypeName(fullName: String) -> String {
+        let fullName = fullName
+            .replacingOccurrences(of: ".keyed: ", with: "")
+            .replacingOccurrences(of: ".single: ", with: "")
```
djones6: I kind of see what's going on here, but I'm not convinced that converting a TypeInfo to a String and then removing the extra text is a good strategy.  Also this ignores other cases (SwaggerGenerator doesn't properly handle dictionaries and arrays nested within each other currently, but we might find `.cyclic` and `.optional` types).

Instead, you could switch over the `TypeInfo` in `getTypeName(typeInfo:)` and extract the value you need.  You could use TypeInfo's [debugDescription](https://github.com/IBM-Swift/TypeDecoder/blob/master/Sources/TypeDecoder/TypeDecoder.swift#L244) as a reference.

123FLO321: Cleaned up the code

123FLO321:
@djones6 Tests added.

djones6:
@123FLO321 This is now tagged in https://github.com/IBM-Swift/Kitura/releases/tag/2.9.1

PR #1491, opened 2019-10-29
Add POST codable route support without input object.

mbarnach:
Add a new codable route for POST with no object as input.

## Description
This allows to create a resource without arguments, like this:
```
router.post("/resource") { (respondWith: (Resource?, RequestError?) -> Void) in
  respondWith(resource, nil)
}
```

## Motivation and Context
It is not possible to use the POST verb without argument, making it weird to create a new resource from an endpoint that has no dependencies.

## How Has This Been Tested?
External testing done. I'm unsure about where to add the test if needed.

## Checklist:
- [X] If applicable, I have updated the documentation accordingly.

### DIFF
--- a/Sources/Kitura/CodableRouter.swift
+++ b/Sources/Kitura/CodableRouter.swift
@@ -271,6 +271,28 @@ extension Router {
         deleteSafely(route, handler: handler)
     }
 
+
+    /**
+     Setup a SimpleCodableClosure on the provided route which will be invoked when a POST request comes to the server.
+     This can be use to create a new resource without parameters.
+
+     ### Usage Example: ###
+     ````
+     //Resource is a struct object that conforms to Codable
+     router.post("/resource") { (respondWith: (Resource?, RequestError?) -> Void) in
+
+        ...
+
+        respondWith(resource, nil)
+     }
+     ````
+     - Parameter route: A String specifying the pattern that needs to be matched, in order for the handler to be invoked.
+     - Parameter handler: A SimpleCodableClosure that gets invoked when a request comes to the server.
+     */
+    public func post<O: Codable>(_ route: String, handler: @escaping SimpleCodableClosure<O>) {
+        postSafely(route, handler: handler)
+    }
+
     /**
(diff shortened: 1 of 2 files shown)


ianpartridge:
Thanks for the PR! We'll get it reviewed.

mbarnach:
I've updated the tests.
For the use case, I guess it could be seen as a semantically different way of saying `GET /resource` but with creation, like for a token, a new "element", etc.
For the REST opinion, I've found that (interesting) thread: [w3 lists](http://lists.w3.org/Archives/Public/ietf-http-wg/2010JulSep/0272.html). Long story short, not so common, but can make sense (so an option to have?). There is also some valid arguments in that [StackOverflow thread](https://stackoverflow.com/questions/4191593/is-it-considered-bad-practice-to-perform-http-post-without-entity-body), especially about the caching involved eventually.

Tests/KituraTests/TestSwaggerGeneration.swift
```diff
@@ -617,6 +623,15 @@ final class TestSwaggerGeneration: KituraTest, KituraTestSuite {
     }
 
     func pathContentAssertions3(paths: [String: Any]) {
+        // Test empty POST
+        if let path = paths["/me/new"] as? [String: Any] {
+            if let new = path["new"] as? [String: Any] {
```
djones6: This line generates a compile warning because `new` is not used.  Could change to `let _ =` or maybe there's a neater way to check the type?

djones6:
@ianpartridge The code for this change looks okay, what do you think?

Tests/KituraTests/TestSwaggerGeneration.swift
mbarnach: Fixed! Thanks.

djones6:
The Travis failure on this job is unrelated to this PR, and has been happening on the CRON builds for master since about 3 weeks ago, eg: https://travis-ci.org/IBM-Swift/Kitura/jobs/648208574
Timeouts in tests when running with the combination of KITURA_NIO on OSX. @harish1992 could you take a look?

lastthyme (approved):

PR #1522, opened 2020-09-23
Improve logging and debugging and make Kitura available for `iOS`

DanielMandea:
## Description
Added support for accessing `serverRequest` under `RouterRequest` in order to improve logging capabilities and debugging. Also added Kitura Nio as the main underlying layer for networking. 

## Motivation and Context
`Kitura Nio`  is mandatory while running on iOS 
While improving logging as part of our internal `RouterMiddleware` that is specialised for logging we were forced at some point to do the following `(request.serverRequest as? HTTPServerRequest)?.buffer?.rewind()` which was impossible due to internal access for `serverRequest`

## How Has This Been Tested?
The change consists in making a constant public instead of keeping it internal. 

## Checklist:


futurejones (commented):
1. Substituting official packages `https://github.com/IBM-Swift/Kitura-NIO.git` with your own personal version `https://github.com/DanielMandea/Kitura-NIO.git` is not acceptable.
If you want changes made to `Kitura-NIO` you need to submit a PR to the `Kitura-NIO` repo and have it approved there. https://github.com/Kitura/Kitura-NIO

2. Removing `https://github.com/IBM-Swift/Kitura-net.git` and forcing everybody to use `Kitura-NIO` is not a good idea. 
If there is an issue caused by defaulting to `Kitura-Net` then maybe the default can be changed to `Kitura-NIO` but leaving the option to use `Kitura-Net`.

mbarnach:
I think you can have the logging behaviour with the `ServerDelegate`.

A pseudo-code like that could work:

```swift
class RequestLogger: ServerDelegate {
    let router: Router
    init(_ router: Router) {
        self.router = router
    }
    func handle(request: ServerRequest, response: ServerResponse) {
        Log.debug("[\(request.remoteAddress)] \(request.method) HTTP \(request.httpVersionMajor ?? 1).\(request.httpVersionMinor ?? 1) \(request.urlURL)")
        router.handle(request: request, response: response)
    }
}
```

And then wrap your router:

```swift
Kitura.addHTTPServer(onPort: 8080, with: RequestLogger(router))
```

This doesn't solve the NIO issue, but hopefully the log issue.

PR #1537, opened 2020-11-01
Fix link to https://www.kitura.dev

kikeenrique:
Fix link that was not working

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,5 @@
 <p align="center">
-<a href="http://kitura.io/">
+<a href="https://www.kitura.dev">
 <img src="https://raw.githubusercontent.com/IBM-Swift/Kitura/master/Documentation/KituraLogo-wide.png" height="100" alt="Kitura">
 </a>
 <p align="center"><strong>A Swift Web Framework and HTTP Server</strong></p>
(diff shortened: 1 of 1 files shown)


dannys42 (approved):

kikeenrique:
As this change is trivial and unlikely to break CI build. Would it be ready to merge?

README.md
```diff
@@ -50,7 +50,7 @@ Kitura is a web framework and web server that is created for web services writte
 
 ## Getting Started
 
-Visit [www.kitura.io](http://www.kitura.io/) for a Getting Started guide, tutorials, and API reference documentation.
+Visit [https://www.kitura.dev](http://www.kitura.dev/) for a Getting Started guide, tutorials, and API reference documentation.
```
dannys42, line 53: Would you mind setting renaming http to https in the text as well?

dannys42:
@kikeenrique Yeah, we'll figure out the CI separately.  Thanks for the update.

README.md
kikeenrique: Done. And I've modified 2 more links.
There is also a link pointing to https://www.kitura.io/packages.html#al, but I don't know where it should point now.

dannys42 (approved):
Thanks!

PR #1543, opened 2021-01-26
Updated github org from IBM-Swift to Kitura; updated jazzy docs

dannys42:


dannys42:
Pushing this through so as to get documentation updated, even though some tests are failing as there is no code change

PR #1549, opened 2021-03-11
Fix test: allow ignoring non-existant file when cleaning up

dannys42:
Fix unit tests

### DIFF
--- a/Tests/KituraTests/KituraTest.swift
+++ b/Tests/KituraTests/KituraTest.swift
@@ -388,9 +388,14 @@ class KituraTest: XCTestCase {
     }
 
     // Delete a temporary file path.
-    func removeTemporaryFilePath(_ path: String) {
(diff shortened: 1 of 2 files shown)


Tests/KituraTests/KituraTest.swift
```diff
@@ -163,18 +163,23 @@ class KituraTest: XCTestCase {
...
                 asyncTask(expectation)
             }
         }
 
         // wait for timeout or for all created expectations to be fulfilled
-        waitForExpectations(timeout: timeout) { error in
-            XCTAssertNil(error)
-        }
+        wait(for: expectations, timeout: timeout)
+//        waitForExpectations(timeout: timeout) { error in
```
mbarnach, line 180: Shall we clean this or do you plan to re-introduce it like that?

PR #1551, opened 2021-04-16
Add an option to reset the buffer before sending.

mbarnach:
Allow to reset the buffer before sending the response.

## Description
When sending a response using `.send`, the body content is inaccessible.
But sometimes, it makes sense to reset the buffer before sending,
eg. an error handler that remove internal details,
sending a JSON payload, not kowing if there is already something inside, etc.

## Motivation and Context
When capturing internal error and making them more explicit, eg. RFC 7807,
we need to reset the buffer to avoid appending internal content
or sending an invalid (JSON) payload.

## How Has This Been Tested?
Add tests to check the resetted content for `String`, `String?` and `Data`
which is used by all underlying methods.

## Checklist:
- [-] If applicable, I have updated the documentation accordingly.

### DIFF
--- a/Sources/Kitura/RouterResponse.swift
+++ b/Sources/Kitura/RouterResponse.swift
@@ -407,18 +407,20 @@ public class RouterResponse {
     /// - Parameter forKey: A value used to match the Encodable values to the correct variable in a template file.
     ///                                 The `forKey` value should match the desired variable in the template file.
     /// - Parameter options: Rendering options, specific per template engine
+    /// - Parameter reset: Reset the body content before sending the response.
     /// - Throws: TemplatingError if no file extension was specified or there is no template engine defined for the extension.
     /// - Returns: This RouterResponse.
     @discardableResult
     public func render<I: Identifier, T: Encodable>(_ resource: String, with values: [(I, T)], forKey key: String,
-                                   options: RenderingOptions = NullRenderingOptions()) throws -> RouterResponse {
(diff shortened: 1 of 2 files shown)


dannys42:
Thanks for the motivation and context!  As an alternative, what do you think about adding a separate function call, like this:

```
\@discardableResult
public func reset() -> RouterResponse {
...
}
```

This would allow you to have syntax like:

```
reset()
send(...)
```

Or:

```
reset().send(...)
```

mbarnach:
It looks like a good idea indeed.
I like the initial syntax until I had to change so many functions... I will give it a try and get back here, either with an update of the PR or my comments if I come across some un-expected changes.

dannys42:
@mbarnach Oops, you're too fast... I had more thoughts. :-)

mbarnach:
@dannys42 I'm working with the following ideas:
- use a dedicated function, like mentioned above;
- call it `reset` to be more generic.
Then, I'm wondering if this function should also reset:
- the `state`;
- the `lifecycle`;
- the cookies;
- the error;
- the `statusCode` (to `.unknown`);
- the `userInfo`
- and even the `headers`.

It seems to be quite a lot, but that would make sense in the case of error reporting.
A solution would then be that we use the parameters for the reset function, in the form of an `OptionSet` where we can define all of these parameters, and expose to the user the ability to reset any of them.

What are your thoughts on that approach?

mbarnach:
@dannys42 A idea of it could be find in #1552

Issue #1556, opened 2021-08-23
No support for Get with ID and Query params


mbarnach:
## Context and Description
Kitura doesn't support GET with ID and query params, like this one:
```http
GET /order/:user_id?page=2
```
This is quite a standard pattern in REST and could beneficial to introduce it inside Kitura.

mbarnach:
@adam-rocska When you have a chance, you can review this PR: https://github.com/Kitura/Kitura/pull/1557 Running for years now for me ;-)

adam-rocska:
I will. I’m still busy with housekeeping steps, but PR cleanups are coming soon.
I just finished shoveling myself through the issues across the org.
Now I’ll do a CI migration round from travis _(or whatever surprises might come up)_ to GitHub Actions, +/- other quality guard contingencies.

Only after these would I dare to touch old dusty PRs. But I will.

PR #1557, opened 2021-08-23
Add the support for ID and Query parameters.

mbarnach:
## Description
Add the support for GET with an ID and optional query parameters.
Supports the return of a single object or an array of objects.
The parameters are only optional in this PR, which is what is making sense most (all?) the time I think.
The return of a list of IDs only has not been done, but could also be added.

Swagger automatic documentation has been added.

## Motivation and Context
Fix issue #1556 that reports the lack of such feature.

## How Has This Been Tested?
A test has been added to the test list.

## Checklist:

### DIFF
--- a/Sources/Kitura/CodableRouter+TypeSafeMiddleware.swift
+++ b/Sources/Kitura/CodableRouter+TypeSafeMiddleware.swift
@@ -64,7 +64,7 @@ extension Router {
             }
         }
     }
-    
+
     /**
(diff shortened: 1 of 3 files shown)


adam-rocska:
Oh my GAWD!

So sorry. If it wasn't for the GitHub widget on my iPad, this I wouldn't have known about this review request...

I assume it's obsolete at this point @mbarnach?

Again: sorry. This is shameful of me...

PR #1558, opened 2021-09-13
Misc ci remove swift4x

dannys42:


Tests/KituraTests/TestServer.swift
```diff
@@ -38,7 +38,7 @@ final class TestServer: KituraTest, KituraTestSuite {
     }
 
     var httpPort = 8080
-    let useNIO = ProcessInfo.processInfo.environment["KITURA_NIO"] != nil
+    let useNIO = ProcessInfo.processInfo.environment["KITURA_NIO"] != "0"
```
mbarnach, line 41: @dannys42 Are you inverting the NIO logic on purpose here or for convenience?

dannys42: Yes, I'm inverting the KITURA_NIO logic to make NIO the default (on our path to deprecating the Kitura-net repo).  This has the advantage of working in iOS and other Xcode projects, which is great for using Kitura as a mock server for unit testing.   I'll update the readme to include this note.

PR #1561, opened 2022-04-11
Swift 5.2 is now the minimum; Update Travis CI to latest docker images

dannys42:
Update dependencies to eliminate some warnings.


dannys42:
@mbarnach Took some effort, but what do you think?

PR #1563, opened 2022-09-09
Make Kitura-NIO the default network stack.

dannys42:
Now use Kitura-NIO rather than Kitura-net.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -21,13 +21,18 @@ import Foundation
 
 var kituraNetPackage: Package.Dependency
 let kituraNetDependency: Target.Dependency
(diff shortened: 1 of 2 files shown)


dannys42:
Alright, going to try this again.  I know the macOS parallel build option is failing... I have no idea why and I think it'll require a full restructuring of the test suite (thanks to the impossibility of debugging the builder pattern).  So I think I'm going to just push this through and release this as a major version update (in case it causes dependency issues for people with other packages).  Any thoughts?

mbarnach (approved):
All good to me 👍 ! Let's do it!

